PR #10, opened 2023-01-27
add decrement test and make it pass

philipcmonk:
Messing around with the code, fixed a few bugs


eamsden (approved):
These are all correct

PR #12, opened 2023-02-02
"implement" "jet" "dashboard"

philipcmonk:
This is an example of the "cheating" option for firing jets, implemented in the treewalking interpreter.  Included are jets for decrement and cut, since those are straightforward to do on bitslices without a separate bignum library.

Additionally, this includes a bugfix in the allocator, basic printing functions for nouns, and renames the axis function to slot because having the verb and noun named the same was awkward.

The commits are self-contained, let me know if you prefer those as separate PRs.


rust/ares/src/mem.rs
```diff
@@ -157,8 +157,8 @@ impl NockStack {
     }
 
     unsafe fn restore_prev_stack_pointer_from_local_west(&mut self, local: usize) {
-        *(self.previous_stack_pointer_pointer_east()) =
-            *(self.slot_pointer_east(local + 2) as *mut *mut u64);
+        *(self.previous_stack_pointer_pointer_west()) =
```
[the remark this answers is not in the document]
eamsden: (this got merged in #13)

rust/ares/src/interpreter.rs
```diff
@@ -505,7 +505,11 @@ fn push_formula(stack: &mut NockStack, formula: Noun) {
     }
 }
 
-fn axis(mut noun: Noun, axis: &BitSlice<u64, Lsb0>) -> Noun {
+pub fn raw_slot(noun: Noun, axis: u64) -> Noun {
```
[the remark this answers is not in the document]
eamsden: This got merged in #13

rust/ares/src/jets.rs
```diff
@@ -0,0 +1,68 @@
...
+pub fn get_jet(jet_name: Noun) -> Result<fn(&mut NockStack, Noun) -> Noun, ()> {
+    if let Ok(jet_direct) = jet_name.as_direct() {
+        match jet_direct.data() {
+            0x636564 => Ok(jet_dec),
+            0x747563 => Ok(jet_cut),
+            _ => Err(()),
+        }
+    } else {
+        Err(())
+    }
```
mcevoypeter, line 15:
```rust
/// Runs a jet.
pub fn exec(jet_name: Noun, stack: &mut NockStack, subject: Noun) -> Result<Noun, ()> {
    match jet_name.as_direct()?.data() {
        0x636564 => Ok(Dec::exec(stack, subject)),
        0x747563 => Ok(Cut::exec(stack, subject)),
        _ => Err(()),
    }
}
```

rust/ares/src/jets.rs
```diff
@@ -0,0 +1,68 @@
...
+                        let res = unsafe { new_indirect.normalize_as_atom() };
+                        res.as_noun()
+                    }
+                }
+            }
+        }
+    } else {
+        panic!("Argument to dec must be atom");
+    }
+}
```
mcevoypeter, line 53:
```rust
struct Dec;

impl Jet for Dec {
    fn exec(_stack: &mut NockStack, _subject: Noun) -> Noun {
        unimplemented!()
    }
}
```

rust/ares/src/jets.rs
```diff
@@ -0,0 +1,68 @@
...
+        match jet_direct.data() {
+            0x636564 => Ok(jet_dec),
+            0x747563 => Ok(jet_cut),
+            _ => Err(()),
+        }
+    } else {
+        Err(())
+    }
+}
+
```
mcevoypeter, line 17: Because all jets implement the same interface, it makes sense to define a `Jet` trait:
```rust
trait Jet {
    /// Runs a jet using the provided subject, returning the result on success and panicking on
    /// failure.
    fn exec(stack: &mut NockStack, subject: Noun) -> Noun;
}
```

rust/ares/src/jets.rs
```diff
@@ -0,0 +1,68 @@
...
+    let run = raw_slot(arg, 13).as_direct().unwrap().data();
+    let atom = raw_slot(arg, 7).as_atom().unwrap();
+    let slice = atom.as_bitslice();
+    let unit = 1 << bloq;
+    let (mut new_indirect, new_slice) =
+        unsafe { IndirectAtom::new_raw_mut_bitslice(stack, ((run * unit + 63) >> 6) as usize) };
+    new_slice[..(unit * run) as usize]
+        .copy_from_bitslice(&slice[(unit * start) as usize..(unit * (start + run)) as usize]);
+    unsafe { new_indirect.normalize_as_atom().as_noun() }
+}
```
mcevoypeter, line 68:
```rust
/// Cut.
struct Cut;

impl Jet for Cut {
    fn exec(_stack: &mut NockStack, _subject: Noun) -> Noun {
        unimplemented!()
    }
}
```

rust/ares/src/interpreter.rs
```diff
@@ -1,4 +1,5 @@
 use self::NockWork::*;
+use crate::jets::get_jet;
```
mcevoypeter, line 2:
```rust
use crate::jets;
```

rust/ares/src/interpreter.rs
```diff
@@ -312,6 +313,20 @@ pub fn interpret(stack: &mut NockStack, mut subject: Noun, formula: Noun) -> Nou
...
+                        .raw_equals(DirectAtom::new_unchecked(0x706f7270).as_noun())
+                    {
+                        if let Ok(jet_formula) = hint_cell.tail().as_cell() {
+                            let jet_name = jet_formula.tail();
+                            let jet = get_jet(jet_name);
+                            if let Ok(jet) = jet {
+                                res = jet(stack, subject);
+                                stack.pop(&mut res);
+                                continue;
+                            }
```
mcevoypeter, line 327:
```rust
if let Ok(mut res) = jets::exec(jet_name, stack, subject) {
    stack.pop(&mut res);
    continue;
}
```

mcevoypeter (commented):
Just some unsolicited Rust comments :) At a high level, I'm of the opinion we should define a `Jet` trait, which defines the interface that any jet implements. Then, for each jet, we define an empty struct, which in Rust parlance is a "zero-sized type", and implement `Jet` on that struct. Because the struct is a zero-sized type, the compiler optimizes it away, yielding the exact same assembly instructions when compared to the function-based approach you took here. With a unique type for each jet, we can also easily implement "marker traits"--traits that don't provide any functionality beyond marking the implementing types as having a certain property--for different jet levels i.e. "absolutely necessary" all the way to "we can turn this jet off without affecting general system performance", which may or not turn out to be useful. In any case, a trait-based approach is more idiomatic, and it'll allow us to enforce more invariants at compile-time via the type system, which is undoubtedly a good thing.

philipcmonk:
Thanks, I appreciate the rust tips!  I've never written rust before, so I don't know shortcuts or idioms.

Can you explain more what a jet trait buys us?  I suppose if we had other operations we wanted to add to a jet, it would keep it all in the same place.

rust/ares/src/jets.rs
```diff
@@ -4,14 +4,10 @@ use crate::noun::{DirectAtom, IndirectAtom, Noun};
 use either::Either::*;
 
 pub fn get_jet(jet_name: Noun) -> Result<fn(&mut NockStack, Noun) -> Noun, ()> {
-    if let Ok(jet_direct) = jet_name.as_direct() {
-        match jet_direct.data() {
-            0x636564 => Ok(jet_dec),
```
eamsden: Looks like we need a macro to convert strings into LE ASCII direct atoms

`tas!("dec")`
`tas!("add")`

rust/ares/src/jets.rs
eamsden: This is only the case for the treewalking interpreter. In the code-generating Nock implementation, jets will specify which subject axes they wish to receive as arguments.

I don't think a trait implementation for this makes things any clearer than the way Philip already wrote it.

mcevoypeter: I still think it makes sense to define a `Jet` trait in the code-generating Nock implementation, even if it requires an associated type. The reason is twofold. One, if a function has an arbitrary jet as a parameter, then we can use the trait to constrain that parameter's type more tightly than if we didn't have a trait. For example:
```rust
/// With a trait.
fn takes_a_jet(jet: impl Jet) { unimplemented!() }

/// Without a trait.
fn takes_a_jet(jet: fn(&mut NockStack, Noun) -> Noun) -> { unimplemented!() }
```
This ensures whoever calling `takes_a_jet()` already has a jet because the compiler won't allow them to pass an arbitrary function whose signature just happens to match.

Two, while I agree the trait implementation doesn't make the code any clearer *at this point* when we only have two jets, when we add more and more jets, the trait will reduce the number of errors a jet-implementer can make. The jet trait clearly communicates exactly what a jet does, explicitly encoding the semantics of a jet in the type system rather than implicitly encoding those semantics in the implementations of jets.

mcevoypeter:
> Can you explain more what a jet trait buys us? I suppose if we had other operations we wanted to add to a jet, it would keep it all in the same place.

https://github.com/urbit/new-mars/pull/12#discussion_r1095936657 may be a somewhat satisfactory explanation. Another way of saying the same thing in fewer words is that a trait is a zero-cost way of explicitly specifying the semantics of jets and making that specification checkable by the type system.

eamsden (approved):

eamsden:
@philipcmonk I'm fine to either merge this and then merge Peter's `tas!` macro after, or to just merge the tas! macro into your branch and then merge. Either way I'd like to get both PRs in

PR #14, opened 2023-02-02
[ares][unifying equality] simplify, document, remove bugs and redundancies

eamsden:
This simplifies the unifying equality implementation, removes some redundant tests, and adds a generous comment explaining the implementation and how it differs from vere.

- Make raw equality the outermost check
- Don't do anything explicity for direct atoms: equal direct atoms will pass the raw equality check: disequal direct atoms will fail it and should be allowed to fall through the type discrimination
- Don't check pointer equality on indirect atoms or cells as that was already done by the raw equality check.

### DIFF
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -622,122 +622,122 @@ impl NockStack {
     }
 }
 
-/** Unifying equality compares nouns for equality in the obvious way, and replaces a noun pointing
- * to a more junior allocation with a noun pointing to a more senior allocation if the two are
- * equal.
+/** Unifying equality compares nouns for equality structurally. It *unifies* noun representations
+ * by rewriting a reference to the more junior (resident in a more recently allocated stack frame)
+ * noun to a reference to the more senior noun, but does so partially and recursively, so as
+ * subnouns are discovered to be equal, they too are unified. Thus even if the overall result is
+ * disequality, some equal subnouns may have been unified.
  *
  * This function is unsafe because it demands that all atoms be normalized: direct and indirect atoms
  * will be considered non-equal without comparing their values, and indirects of different sizes
  * will be considered non-equal.
- *
- * TODO: we really should try to tie lifetimes into the stack and use mut references instead of raw
- * pointers wherever we can, but this is hard and delaying progress right now
  */
 pub unsafe fn unifying_equality(stack: &mut NockStack, a: *mut Noun, b: *mut Noun) -> bool {
+    /* This version of unifying equality is not like that of vere.
+     * Vere does a tree comparison (accelerated by pointer equality and short-circuited by mug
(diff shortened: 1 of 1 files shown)


rust/ares/src/mem.rs
```diff
@@ -622,122 +622,122 @@ impl NockStack {
...
+     * respectively, of cells, and then re-testing. If unification succeeds then the heads and
+     * tails will be pointer-wise equal and the cell itself can be unified. A failed unification of
+     * the head or the tail will already short-circuit the unification/equality test, so we will
+     * not return to re-test the pointer equality.
+     *
+     * When actually mutating references for unification, we must be careful to respect seniority.
+     * A reference to a more junior noun should always be replaced with a reference to a more
+     * senior noun, *never vice versa*, to avoid introducing references from more senior frames
+     * into more junior frames, which would result in incorrect operation of the copier.
+     */
```
philipcmonk, line 658: I suspect it would be worth doing a raw_equals at the very beginning of the function, before even allocating a stack frame, because that case is probably pretty common.  Maybe even a mug-check, because it's quite common for the mug cache to exist.

rust/ares/src/mem.rs
```diff
@@ -622,122 +622,122 @@ impl NockStack {
...
-                    if x_direct.data() == y_direct.data() {
+            match (x_alloc.as_either(), y_alloc.as_either()) {
+                (Left(x_indirect), Left(y_indirect)) => {
+                    let x_as_ptr = x_indirect.to_raw_pointer();
+                    let y_as_ptr = y_indirect.to_raw_pointer();
+                    if x_indirect.size() == y_indirect.size()
+                        && memcmp(
+                            x_indirect.data_pointer() as *const c_void,
+                            y_indirect.data_pointer() as *const c_void,
+                            indirect_raw_size(x_indirect),
```
philipcmonk, line 690: This is a word size, right?  Does memcmp take bytes?

rust/ares/src/mem.rs
eamsden: I actually tried this and it made millisecond-or-less difference for jam.

rust/ares/src/mem.rs
eamsden: Woops yup good catch. `indirect_raw_size` adds the 2 extra words for the metadata and size fields but does *not* convert to byte size, and `memcmp` is a pretty raw binding to libc so it does take bytes.

eamsden:
```suggestion
                            indirect_raw_size(x_indirect) << 3,
```

philipcmonk (approved):

rust/ares/src/mem.rs
eamsden: Actually once the bug below was fixed this started making a difference so those checks are in now.

PR #16, opened 2023-02-03
[ares] fix atom initialization bug

philipcmonk:
We already added 2 to the buffer in new_raw_mut, no need to do it again.  This fixed a bug in something I was tinkering with.


eamsden (approved):
Good catch

PR #20, opened 2023-02-11
[ares] convenient noun constructors

philipcmonk:
I have found these constructors convenient, otherwise writing out even small nouns is very long.  Arguably `new_tuple` should be `C` or something.

### DIFF
--- a/rust/ares/src/noun.rs
+++ b/rust/ares/src/noun.rs
@@ -150,6 +150,11 @@ impl fmt::Debug for DirectAtom {
     }
 }
 
+#[allow(non_snake_case)]
+pub fn D(n: u64) -> Noun {
+    DirectAtom::new_panic(n).as_noun()
+}
+
 /** An indirect atom.
  *
  * Indirect atoms represent atoms above DIRECT_MAX as a tagged pointer to a memory buffer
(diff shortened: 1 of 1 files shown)


rust/ares/src/noun.rs
```diff
@@ -387,6 +392,19 @@ impl Cell {
         }
     }
 
+    pub fn new_tuple(allocator: &mut dyn NounAllocator, tup: &[Noun]) -> Cell {
```
eamsden, line 395: maybe call this `T` since it constructs a conventional noun tuple.

eamsden:
```suggestion
    pub fn T(allocator: &mut dyn NounAllocator, tup: &[Noun]) -> Cell {
```

rust/ares/src/noun.rs
```diff
@@ -150,6 +150,11 @@ impl fmt::Debug for DirectAtom {
     }
 }
 
+#[allow(non_snake_case)]
+pub fn D(n: u64) -> Noun {
```
eamsden, line 154:
```suggestion
pub const fn D(n: u64) -> Noun {
```

This signals to the compiler that this function can be evaluated at compile time.

eamsden (changes requested):
A better name for `new_tuple` is suggested but up to you. The `D` smart constructor should be marked `const` to enable compile-time evaluation when it's called with a constant (as it quite often will be.) Otherwise this is excellent.

PR #21, opened 2023-02-11
[ares] add basic ipc handler

philipcmonk:
This is a pretty minimal implementation of the newt protocol used by the king.  It works, in the sense that the vere king starts up with this serf process and tries to replay, then passes in events, which we can print out here.  Obviously, we don't actually run the events, we just report success.


philipcmonk:
Added another commit which expands the functionality to actually run events.  With the included "baby" pill (which is a degenerate "arvo" that requires no jets), I was able to boot and handle events with vere's king.

rust/ares/src/interpreter.rs
```diff
@@ -64,7 +65,12 @@ fn noun_to_work(noun: Noun) -> NockWork {
     }
 }
 
-pub fn interpret(stack: &mut NockStack, mut subject: Noun, formula: Noun) -> Noun {
+pub fn interpret(
+    stack: &mut NockStack,
+    newt: &mut Option<&mut Newt>,
```
eamsden, line 70: It would be good to have a comment regarding the purpose of the `newt` argument and the expected behavior in the `Some` and `None` cases.

rust/ares/src/interpreter.rs
```diff
@@ -314,30 +320,27 @@ pub fn interpret(stack: &mut NockStack, mut subject: Noun, formula: Noun) -> Nou
...
-                    {
-                        if let Ok(jet_formula) = hint_cell.tail().as_cell() {
-                            let jet_name = jet_formula.tail();
-                            if let Ok(jet) = jets::get_jet(jet_name) {
-                                res = jet(stack, subject);
-                                stack.pop(&mut res);
-                                continue;
-                            }
-                        }
+                    if let Ok(found) = match_pre_hint(stack, subject, hint_cell) {
```
eamsden, line 323: Can we have a comment explaining the `pre_` vs `post_` distinction for hint matching? I figured it out but it took a bit. Even something like `hint matching, tree preorder` vs `hint matching, tree postorder`

rust/ares/src/newt.rs
```diff
@@ -0,0 +1,231 @@
...
+                copy_nonoverlapping(
+                    &direct.data() as *const u64 as *const u8,
+                    buf.as_mut_ptr().add(5) as *mut u8,
+                    size,
+                );
+            },
+            Either::Right(indirect) => unsafe {
+                // REVIEW: is this safe/the right way to do this?
+                copy_nonoverlapping(
+                    indirect.data_pointer() as *const u8,
```
eamsden, line 99: Does the data need to be copied out of the indirect atom to be written? Could we just write the size from the buffer and then the data directly from the indirect atom's data pointer?

rust/ares/src/newt.rs
```diff
@@ -0,0 +1,231 @@
...
+                return None;
+            } else {
+                panic!("Error reading header: {}", err);
+            }
+        }
+
+        let byte_len = u32::from_le_bytes([header[1], header[2], header[3], header[4]]) as usize;
+
+        // Would be nice to copy directly into an indirect atom, but I don't know how to do that
+        // when these are aligned to bytes
```
eamsden, line 210: Have a look at what `jam()` does.

eamsden (changes requested):

PR #24, opened 2023-02-14
[ares] fix a couple bugs in unifying equality

philipcmonk:
East/West typo, also we were setting the frame pointer to the stack pointer while chasing up the stack.

This also refactors the conditionals in a way that I think is easier to read.  We always have to do both calculations anyway.

### DIFF
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -777,21 +777,14 @@ unsafe fn senior_pointer_first<T>(
         ),
     };
     loop {
-        if a < high_pointer && a >= low_pointer {
-            // a is in the current frame
-            if b < high_pointer && b >= low_pointer {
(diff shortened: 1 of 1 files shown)


rust/ares/src/mem.rs
```diff
@@ -777,21 +777,14 @@ unsafe fn senior_pointer_first<T>(
...
-                // b is not, so b must be further up, b is senior
-                break (b, a);
-            }
-        } else {
-            // a is not in the current frame
-            if b < high_pointer && b >= low_pointer {
-                // b is, a is senior
-                break (a, b);
-            } else {
+        match (
```
eamsden, line 780: Definitely much easier to read, thank you!

eamsden (approved):

PR #25, opened 2023-02-14
Math jets

philipcmonk:
Here's all the /a jets and a few of the /c jets, implemented with ibig.  Further tasks:
- [x] Finish the math jets
- [x] Patch ibig to use atoms directly to avoid all copying and ensure it uses the nock stack.  Internally, its data type is almost exactly the same as our atoms, so this shouldn't be too hard.  With luck, this will be faster than gmp for most things because we won't need to copy
- [x] Run the L2 history (which is pretty atom-jet heavy) as a realer benchmark against vere.

(Also a fix in unifying equality: f550bbb43a47add507637b7eb0bd875123058982)


eamsden:
@philipcmonk please cherry pick f550bbb to a new PR so we can get that merged right away

philipcmonk:
Added the remainder of the atom jets in lib/tiny, and added a few unit tests for each jet.

I moved `/jets_math.rs` into `/jets/math.rs` (ie into the jets module).  I'm given to understand that it's customary to keep unit tests in the same file as the functions, but that it's also permissible to put them inside the module in a separate file (ie in `/jets/tests.rs`, `/jets/math_tests.rs` or even `/jets/math/tests.rs`).  Right now they're in the same file, but I wouldn't be opposed to moving them.

philipcmonk:
In the last commit, I extracted a bunch of little changes that I made while debugging, and also included the source for the pills I've been using.  I don't include the compiled pill, but I could if that's helpful.

Depending how this goes, some of the changes there could be upstreamed earlier.  Alternately, we could just merge this PR with the ibig changes as a TODO, so that we're working off the same branch most of the time.  I have a slight preference for this approach.

PR #29, opened 2023-02-16
HAMT, and trait-based object preservation on stack pop

eamsden:


eamsden:
@philipcmonk I rewrote the HAMT to be much clearer and less raw. Maybe we can go over it tomorrow?

@mcevoypeter I would appreciate your review of the HAMT code in particular

rust/ares/src/hamt.rs
```diff
@@ -0,0 +1,407 @@
...
+                            // if we haven't hit depth limit yet we shouldn't be chaining
+                            // we'll make a fake node pointing to the old leaf and "insert into" that
+                            // next time around
+                            assert!(leaf.len == 1);
+                            let fake_buffer = stack.struct_alloc(1);
+                            *fake_buffer = Entry {
+                                leaf: leaf
+                            };
+                            let next_stem = Stem {
+                                bitmap: chunk_to_bit(chunk),
```
eamsden, line 244: I think this is a bug. This should be the next chunk out of the mug since we'll "insert" into this fake node one level down, next time around.

eamsden:
Latest commit adds a mutable HAMT, since using the immutable HAMT for many insertions in a single NockStack frame (as `jam()` and `cue()` do) will result in a lot of garbage without any deallocations. The mutable HAMT takes the same design choice as vere: preallocating 32-slot nodes and directly indexing them with chunks. The only discarded allocations on an insert are leafs, when a new entry must be inserted at an existing leaf. This could be mitigated by giving leaves preallocated capacity and doubling it whenever reallocation became necessary.

eamsden:
@philipcmonk this is ready for review now

philipcmonk (approved):

PR #30, opened 2023-02-23
Use patched ibig to do all allocations on nock stack

philipcmonk:
This pairs with https://github.com/tczajka/ibig-rs/compare/main...urbit:ibig-rs:main to patch ibig-rs to use the nock stack for all allocations (for the operations we use).  The patch to ibig-rs is fairly invasive, but I don't know an easier way.

This includes the assert-no-alloc crate and ties it to the interpreter and jet tests, so that it will crash if we allocate at times when we're not supposed to.  This was useful for testing, and I think we should leave it on, since it gets compiled out of releases anyway.  If this crash is annoying you (eg because your printf allocates), just comment out the custom allocator in lib.rs.

This is not quite optimal for the number of copies from atoms to ubig and vice versa, but it's better than before.  Before, going from ubig to an atom required two copies: one to Vec, then another to an atom.  Now, we produce a Slice, so only the second copy happens.  Going from an atom to a ubig always copies once.  This is less expensive than it sounds, because frequently the left operand is modified in-place.

I don't have tests that thoroughly exercise the performance of bignum math, but the azimuth pill takes approximately the same amount of time to process.  If bignum math becomes a hot spot, it could be worth a more invasive patch to ibig-rs to operate direclty on atoms.

(This isn't based off the merge of philip/jets with eamsden/debug-immutable-hamt because one of my tests doesn't work due to a problem in MutHamt that I haven't debugged)

### DIFF
--- a/rust/ares/Cargo.lock
+++ b/rust/ares/Cargo.lock
@@ -13,6 +13,7 @@ name = "ares"
 version = "0.1.0"
 dependencies = [
  "ares_macros",
+ "assert_no_alloc",
  "bitvec",
  "criterion",
  "either",
@@ -32,6 +33,12 @@ dependencies = [
  "quote",
  "syn",
 ]
+
+[[package]]
+name = "assert_no_alloc"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "55ca83137a482d61d916ceb1eba52a684f98004f18e0cafea230fe5579c178a3"
 
 [[package]]
 name = "atty"
--- a/rust/ares/Cargo.lock
+++ b/rust/ares/Cargo.lock
@@ -265,8 +265,7 @@ dependencies = [
 [[package]]
 name = "ibig"
 version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


philipcmonk:
That is still significantly more invasive. Eg ibig does a lot to try and modify one of the inputs instead of reallocating whenever possible, which is the opposite of what we need. Can absolutely be done, but I think the priority is just a lot lower since it's (non-bottleneck) performance vs this at least gets us memory discipline.

eamsden:
That makes sense.

eamsden:
One more thought: since this is a really heavy vendoring of ibig that we are unlikely to merge back upstream, should we just put the vendored ibig in our tree?

philipcmonk:
Would be convenient.  OTOH, keeping it separate (or as a submodule) might make it easier to integrate upstream changes as they come out?

eamsden:
I wonder how frequently there will be upstream changes that we need to merge? The functionality is what we want as it stands. Mergeable upstream changes would take the form of either performance improvements or bugfixes.

philipcmonk:
Yeah, not very many.  There's not a lot of changes in the last year, for example.

PR #37, opened 2023-03-05
Add CI for Hoon code

### LINKED ISSUE #34: CI: Add MacOS and Windows build flows
See [this advice](https://github.com/urbit/new-mars/pull/31#discussion_r1120287776) from @mcevoypeter

ashelkovnykov:
Resolves #34 

Duplicates the existing workflow for Ares to run on pull requests as well.

Adds a CI workflow for the Hoon code that runs on merges and pull requests which will:
- launch a fake ship
- build the Hoon code
- runs tests


eamsden (changes requested):
Looks like the "success" is a faulty indication:

(from the build log)

image

The `kill` command is printing out usage information and then exiting with success. Meanwhile the build thread is not running.

ashelkovnykov:
@eamsden
The actual issue is that it's calling the cleanup script after a successful run, and that the script is failing to correctly clean up.

ashelkovnykov:
With regards to the thread failure, that's because there's no tests yet. If there were Hoon unit tests copied into the folder, they would be run. If we were using `conn.c`, we could easily detect success/failure.

I added the action to the workflow just so that it already exists, and offers some information (even though it cannot fail in a blocking way). If you'd prefer, I can remove it and it can be re-added later once there are Hoon tests in New Mars and a client for interfacing with `conn.c` is available.

eamsden:
@ashelkovnykov What I want for now is something that will block merging if one of the `hoon/scaffolding/*.hoon` files fails to compile.

ashelkovnykov:
@eamsden This'll do it - there's just also a test action appended to the workflow that doesn't do anything yet (since there's no Hoon tests).

For some reason, the `curl` command to `%lens` for this test job sometimes takes as much as 10 minutes to complete. I have no idea why - whenever I tested it in any combination locally, it worked fine, but as soon as it gets on a runner it takes forever. If you want, I can turn it off until there's actual unit tests and the thin client is released.

eamsden:
@ashelkovnykov it's fine to leave it, 10 minutes is not horrible in a CI context. It would be worth debugging it at some point though.

eamsden:
@ashelkovnykov can you link here to a build failing for failure to compile Hoon so reviewers can verify?

ashelkovnykov:
@eamsden I added a commit that changes line `14` of `hoon/scaffolding/azimuth-pill` to `?~  bogs`, after which the CI build [fails to compile](https://github.com/urbit/new-mars/actions/runs/4345427281/jobs/7590120120). You can see that the build just prior to that one [succeeded](https://github.com/urbit/new-mars/actions/runs/4341935473/jobs/7582142541).

Note that the logs tell you which file failed, but not why. So long as we're using `%lens` instead of `conn.c` + Khan, we're not going to get any better info. That is to say, after the thin client is published, we should do a second pass over these CI flows.

The above applies for the tests job as well. There's nothing to gain from debugging `%lens`; we should just live with it until the better alternative is available.

eamsden (approved):

eamsden:
@ashelkovnykov please revert the failing hoon commit so we have green CI, then we can merge

PR #38, opened 2023-03-08
[ska] prototypes for initialization

eamsden:
This fixes a bug in `+mous` (sock intersection) which was causing infinite loops, and adds lib/sky.hoon which contains prototypes:
- `+dope` which demonstrates simply masking out axes
- `+wove` which translates hoon types into subject knowledge
- `+arid` which uses a fixed point of evaluation and knowledge intersection to find the part of a core which is constant across state transitions.


philipcmonk:
This will need to be reviewed synchronously, happy to do that this week.

philipcmonk (approved):
We walked through this synchronously and made a couple small changes.  Looks good to me now

PR #39, opened 2023-03-10
Address (most) lints suggested by `cargo clippy`

mcevoypeter:
This PR cleans up (most of) the code flagged by Rust's linter, `cargo clippy`. A large number of `warning: unsafe function's docs miss `# Safety` section` lints remain, which we should address. Reasoning about undefined behavior in Rust is significantly harder than in C since safe Rust makes so many more assumptions than C does; unsafe Rust code, therefore, must be aware of the implications of its unsafe behavior for safe Rust code. Documenting all `unsafe` blocks with a quick justification of why the particular `unsafe` block is in fact safe and why it's necessary will make it easier to understand the code and to track down bugs related to `unsafe` blocks.

### DIFF
--- a/.github/workflows/ares-shared.yml
+++ b/.github/workflows/ares-shared.yml
@@ -21,7 +21,7 @@ jobs:
         run: cargo fmt --check
 
       - name: Lint
-        run: cargo clippy -- --deny warnings
+        run: cargo clippy -- --deny warnings --allow clippy::missing_safety_doc
 
       - name: Build
         run: cargo build --verbose
--- a/.github/workflows/ares-shared.yml
+++ b/.github/workflows/ares-shared.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Format
         run: cargo fmt --check
 
+      - name: Lint
+        run: cargo clippy -- --deny warnings
+
       - name: Build
         run: cargo build --verbose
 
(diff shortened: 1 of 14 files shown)


eamsden:
If we're going to do this let's do it *once* and make clippy a blocking CI job so we don't have multiple can't-revert-past diff points in the code

philipcmonk (commented):
I didn't have clippy installed; in line with the rest of the rust toolchain this should be added to `/ares/rust/shell.nix`.  It was as simple as adding `"clippy"` for me.

Apart from that, I feel like these changes are broadly good, and the included CI requirement that cargo clippy pass is a good thing.

It looks like the requirement to document unsafe blocks is specifically for unsafe public functions, which is less than I was initially imagining.  It looks like the rule is "is there a doc comment with a markdown header labeled 'Safety'?", so the minimum is adding `/** # Safety */`, though normally I would expect them to look like:

```
/** <other docs>
 *
 * # Safety
 *
 * Unsafe due to direct pointer use
 */
```

I'm not opposed to adding some basic info to those functions.  However, if we don't want to do that (or not yet), we should still enforce the rest soon, so that we avoid having larger conflicts down the road.

mcevoypeter:
Thanks for pointing out the issue with `clippy` and Nix; I added `clippy` to `shell.nix` as you suggested, which also gives me access to `clippy` within `nix-shell`.

Good point about wanting to enforce the non safety comment lints sooner rather than later. To that end, I disabled the `missing_safety_doc` lint in CI. We can revisit its usage later, especially because it won't cause conflicts regardless of what we decide.

philipcmonk:
Great, then IMO this is good to go now.  @eamsden may have further thoughts

eamsden (approved):
LGTM

PR #40, opened 2023-03-14
Replace dynamic trait objects with static type parameters

mcevoypeter:
This PR replaces dynamic trait objects with static type parameters to eliminate the overhead of a vtable lookup at runtime. It does by requiring types implementing the `NounAllocator` or `Stack` traits also implement the [`Sized`](https://doc.rust-lang.org/std/marker/trait.Sized.html) trait. This change does not result in an increase in binary size:
```
$ git checkout status
$ cargo build --release
$ du -sh target/release/libares.rlib
1.1M    target/release/libares.rlib

$ git checkout peter/remove-dyn
$ cargo build --release
$ du -sh target/release/libares.rlib
1.1M    target/release/libares.rlib
```

### DIFF
--- a/rust/ares/src/noun.rs
+++ b/rust/ares/src/noun.rs
@@ -3,10 +3,9 @@ use bitvec::prelude::{BitSlice, Lsb0};
 use either::Either;
 use ibig::{Stack, UBig};
 use intmap::IntMap;
-use std::{
-    fmt, ptr,
-    slice::{from_raw_parts, from_raw_parts_mut},
-};
+use std::fmt;
+use std::ptr;
+use std::slice::{from_raw_parts, from_raw_parts_mut};
 
 crate::gdb!();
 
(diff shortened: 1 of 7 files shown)


rust/ares/src/noun.rs
```diff
@@ -3,9 +3,10 @@ use bitvec::prelude::{BitSlice, Lsb0};
 use either::Either;
 use ibig::{Stack, UBig};
 use intmap::IntMap;
-use std::fmt;
-use std::ptr;
-use std::slice::{from_raw_parts, from_raw_parts_mut};
+use std::{
```
eamsden, line 6: Please do not mix formatting and semantic changes

eamsden (approved):
Aside from separating formatting/semantic changes (de minimis in this case) this looks good.

eamsden:
@mcevoypeter go ahead and merge this

PR #44, opened 2023-03-28
Add rev jet

h33p:
Hello! Finally got time to finish this up.

- Added jet for `rev`.
- Added utility functions for mutable access to noun bits.
- Rewrote `from_ubig` to not have that long chain of bitshifts, as I was bored on the plane back home.

I don't consider this fully done. Most notably:

- Tests need to be improved.
- Is it okay for nouns to have leading zero bytes? This could come up with direct manipulation to noun's bits, and mutable access functions allow for that.
- Should this go to a different module?

One thing I wanted to ask, is there a nice way to generate test cases? I would imagine basing expectations on known good implementation (vere) would be a good idea.

### DIFF
--- a/rust/ares/src/jets/math.rs
+++ b/rust/ares/src/jets/math.rs
@@ -651,6 +651,41 @@ pub fn met(bloq: usize, a: Atom) -> usize {
     }
 }
 
+pub fn jet_rev(stack: &mut NockStack, subject: Noun) -> Result<Noun, JetErr> {
+    let arg = raw_slot(subject, 6);
+    let boz = raw_slot(arg, 2).as_atom()?.as_direct()?.data();
+
+    if boz >= 64 {
+        return Err(NonDeterministic);
+    }
+
+    let boz = boz as usize;
+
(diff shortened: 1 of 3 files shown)


rust/ares/src/jets/math.rs
```diff
@@ -651,6 +651,41 @@ pub fn met(bloq: usize, a: Atom) -> usize {
...
+
+    let boz = boz as usize;
+
+    let len = raw_slot(arg, 6).as_atom()?.as_direct()?.data();
+
+    let dat = raw_slot(arg, 7).as_atom()?;
+
+    let bits = len << boz;
+
+    let mut output = if dat.is_direct() && bits < 64 {
```
eamsden, line 670: I would appreciate a comment here noting that the `< 64` constraint is due not to word size but to the maximum allowable bits in a direct atom.

eamsden (approved):
Requested a comment otherwise LGTM

rust/ares/src/jets/math.rs
h33p: Added the comment!

PR #52, opened 2023-05-08
Move PMA code into Ares from separate repo

ashelkovnykov:
This PR moves the PMA code, and therefore the active development of the PMA, into Ares. The code is synced from [ashelkovnykov/pma_malloc, commit hash 3329657abccdd06a65d21395e338343289c99ce9](https://github.com/ashelkovnykov/pma_malloc/tree/3329657abccdd06a65d21395e338343289c99ce9).

I have replaced the existing `README` with a temporary one which will be updated once the final design of the PMA is in the code, since a great deal of the current `README` is an explanation of why I used Arrays instead of B Trees.

Note that this PR is *before* applying Phil's patch, therefore there is another PR for the PMA coming shortly.

Some questions:
- What should be done about the licenses? LMDB is OpenLDAP and `phk_malloc` is "beerware". How does this square with MIT?
- I've kept the structure of the PMA code almost identical to what it was in the original repo, but I'm open to suggestions on how to move things around (specifically, the makefile and tests).

@eamsden 
@barter-simsum 
@philipcmonk

### DIFF
--- a/rust/ares/src/pma/.gitignore
+++ b/rust/ares/src/pma/.gitignore
@@ -1,0 +1,3 @@
+bin/
+dep/
+obj/
--- a/rust/ares/src/pma/README.md
+++ b/rust/ares/src/pma/README.md
@@ -1,0 +1,8 @@
+## PMA - TODO
+
+Ported from development in a
+[separate repo](https://github.com/ashelkovnykov/pma_malloc). README will be
+updated after the final implementation is complete, which replaces the
+array-based page directory with a B+ Tree one. Until then, please refer to the
+README in the above-linked directory.
+
--- a/rust/ares/src/pma/includes/checksum.c
+++ b/rust/ares/src/pma/includes/checksum.c
@@ -1,0 +1,135 @@
+/*
+ * Library: libcrc
+ * File:    src/crc32.c (herein src/includes/checksum.c)
+ * Author:  Lammert Bies
+ *
+ * This file is licensed under the MIT License as stated below
+ *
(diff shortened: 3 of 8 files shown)


eamsden (approved):

eamsden:
@ashelkovnykov sorry I missed your questions before merging!

Is the code in here derivative of LMDB and/or phk_malloc?

ashelkovnykov:
@eamsden That seems like a legal question that I'm not qualified to answer. Is there a part of the code that can be explicitly pointed at and said, "This was copied from PHK `malloc` / LMDB"? No. Are there parts of the code that are reminiscent of those projects? Yes, though they have been modified to suit our purposes. Did we openly borrow ideas from those projects for our design? Yes.

philipcmonk:
Openly borrowing ideas is no problem at all -- there are no patents involved.  Writing new code without essentially copy-and-paste is unlikely to be an issue either, especially given the substantial differences between those projects and ours.  So, I believe we have complete copyright over the code we wrote and can license it as we see fit (MIT).

Even if somehow it was determined that this was sufficiently derivative of those projects as to give them copyright protection, OpenLDAP's license and beerware are substantially the same as MIT, so the most we would have to do in that case would be include their license statements somewhere.

So, my opinion is that we can license everything MIT, and the repo is already licensed MIT at the top level, so this is fine.

I note in this PR that out of an abundance of caution we include the beerware license in pma.c.  This seems fine as well.  Ideally we would additionally include a notice saying that much of the code is written by Alex/Tlon, so that it doesn't look like phk is responsible for bugs we introduced.

PR #54, opened 2023-05-12
SACK: Subjective Analysis of Knowledge and Constraints

eamsden:
This PR upgrades SKA to SACK, which is capable of discerning which knowledge is actually necessary to fix called nock formulas for direct calls, and use this information as the battery label for generated code.

This avoids the TMI problem when using SKA to build cell structures in accumulators, and the battery discovery problem which required several ad-hoc heuristics to discover stable batteries for dynamic cores such as Arvo, vanes, gall agents, threads, generators, and dojo expressions. It also allows us to directly derive "cold state"[1] for jetting from fast hints, and to statically link jets using no more information than is already used for the current vere jetting system.

**jetting**
[1]the cold state is the reification of the Nock-side declaration that a core inhabits a certain place in a hierarchy of named cores. It records a battery or batteries for a named core along with a link to the core's parent core or a typed version constant (usually the Hoon kelvin version). The hot state is the set of available jets in the current runtime, annotated by which path into the hierarchy of core names they match. The warm state is the mapping between cold and hot states which allows jets to be executed. So in this formulation of codegen, the cold state is a map of paths into the named core hierarchy to socks which represent the batteries of an entire lineage of cores. The hot state is compiled into ares (or perhaps augmented by jetpacks) and maps paths into the named core hierarchy to jets. The warm state then is the segment of the codegen table (which is labeled by subject-sock+formula pairs) where the "generated" code is actually jets rather than code compiled from nock, and is generated by making, for each jet in the hot state, an entry into the codegen table corresponding to the battery recorded in the cold state for the same path.

_This PR is a draft and exists for visibility. It is not yet code complete, let alone tested._


eamsden:
current testing instructions:
- sync hoon/codegen to a desk containing base-dev e.g. %sandbox
- `~zod:dojo> =sack -build-file /=sandbox=/lib/sack/hoon`
- `~zod:dojo> =deco =>  %none  |%  ++  dec  |=(a=@ ?<(=(a 0) =|(b=@ |-(`@`?:(=(a .+(b)) b $(b .+(b)))))))  --`
- `~zod:dojo> (rout:sack [[& .] !=((dec 5))]:deco)`

eamsden:
@joemfb @frodwith @philipcmonk PTAL, test instructions for current state of work in above thread

eamsden:
Superseded by #89

Issue #50, opened 2023-05-05
2stackz: use split stack frames


eamsden:
https://github.com/urbit/ares/blob/7176f39cf725ff7f7678633abaf7380906ef19bc/rust/ares/src/mem.rs#L34-L66

PR #60, opened 2023-06-15
split stack

drbeefsupreme:
Addresses #50 

Following @eamsden's advice and a couple attempts so far at refactoring the old version, I am rewriting `mem.rs` from scratch since this change touches most of the file.


rust/ares/src/mem.rs
```diff
@@ -24,86 +31,59 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
 }
 
 /** Which side of the two opposing stacks are we working on? */
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, Debug)]
 pub enum Polarity {
```
eamsden, line 35: This should go away

drbeefsupreme: https://github.com/urbit/ares/pull/60/commits/df5fdafe9cc86a2182ef4ce36f7104fb70d78a4a

So far I think it makes the code a bit uglier, but we do get the benefit of a `panic!` in an obvious spot.

Actually, I suppose I can still use `match` statements here, which are easier for me to read than the `if` statements. Will make that change

drbeefsupreme: https://github.com/urbit/ares/pull/60/commits/c4bd517676f569bf0712608d22445b6e22ecdf2f

drbeefsupreme:
In order to not make @eamsden navigate through a huge, constantly changing mess, here's how I plan to iterate:

Aspirationally, push one batch of commits every day or two, and don't push more commits until Ed has time to look them over and leave feedback. Only include functions with a `todo!()` that I'm reasonably sure actually need to exist. Try to flesh out `todo!()`s) in terms of dependence from a root task (allocation, copying, pushing, etc).

So I'm marking https://github.com/urbit/ares/pull/60/commits/0b4029a005842484f18ec47c633389d5a4434761 as my last commit I'll be pushing from this batch. Feedback I'm hoping to get:
 - answers to questions listed in ``//TODO` comments
 - pointing out functions that just have a `todo!()` in them that I shouldn't end up needing (based on their usage pre-split frame)
 - check the logic on initialization (`NockStack::new()`), pushing (`push()`), and allocation (`raw_alloc()`, `indirect_alloc()`, `struct_alloc()`)
 - style and conventions

rust/ares/src/mem.rs
```diff
@@ -23,102 +30,105 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
+    // is in range, then check to see if that is a legitimate pointer or
+    // if its pointing to the middle of an allocated object
     #[inline]
-    pub fn in_frame<T>(&self, ptr: *const T) -> bool {
+    pub unsafe fn in_frame<T>(&self, ptr: *const T) -> bool {
         let ptr_u64 = ptr as *const u64;
-        match &self.polarity {
-            Polarity::East => ptr_u64 >= self.stack_pointer && ptr_u64 < self.frame_pointer,
-            Polarity::West => ptr_u64 >= self.frame_pointer && ptr_u64 < self.stack_pointer,
+        match self.is_west() {
```
eamsden, line 125: Please use if/else instead of match on bools

rust/ares/src/mem.rs
eamsden: > [df5fdaf](https://github.com/urbit/ares/commit/df5fdafe9cc86a2182ef4ce36f7104fb70d78a4a)
> 
> So far I think it makes the code a bit uglier, but we do get the benefit of a `panic!` in an obvious spot.
> 
> Actually, I suppose I can still use `match` statements here, which are easier for me to read than the `if` statements. Will make that change

match on boolean is not good style, please do not do this

rust/ares/src/mem.rs
drbeefsupreme: I think it looks uglier that way but I concede that its much more standard. Will do.

drbeefsupreme:
Another batch of commits.

I'm mostly looking for comments on `copy_west()`. The `pre_copy()` stuff won't be fully wired up until I write `pop_west()` and `pop_east()`, and might need to change depending on where the lightweight stack actually goes (I think at least for copying, we can't put it next to the FP-relative slots as suggested in @eamsden's split stack writeup - but I think it might be fine to put it next to the allocations).

drbeefsupreme:
@eamsden

Based on what we talked about, it seems we do actually want a 3rd pointer in `NockStack`. The last two commits add the `stack_pointer` and then I rewrote `copy_west()` taking this into account. This also "steals" `stack_pointer` for the lightweight stack for traversing nouns being copied. Knowing that `stack_pointer` is the one and only pointer used for lightweight stacks might make things less confusing (we could have introduced a new local variable pointer instead), but it might also create confusion since we're moving the lightweight stack from (in the west case) the high edge of low memory to the low edge of high memory. But once we're at the copying phase, the frame is being destroyed anyways, so maybe that's fine.

It is safe to ignore the `copy_west()` from yesterday.

drbeefsupreme:
@eamsden Here's the new `copy_west()` and `senior_pointer_first()` that treats the FP-relative slots correctly (in the previous version, they were not adjacent to the frame pointer) and refactors slot accessing to be a bit more readable using magic numbers.

drbeefsupreme:
@eamsden here's my first attempt at `unifying_equality()`

It may be evident that I still have some confusion over Rust's raw pointer interface (you can see me casting the same pointer to both `*mut u64` and `*mut (*mut Noun, *mut Noun)`). Here's what I attempted to do:

- Unlike the currently implemented `unifying_equality()`, we do not need to push a new frame to do this
- Allocations for `unifying_equality()` are done in the current frame's allocation arena. We push pointers to these allocations onto the lightweight stack (using `stack_pointer`).
- The loop is declared over when the stack is empty (`frame_pointer == stack_pointer`). However I don't know if `unifying_equality()` might be called when the lightweight stack wasn't already empty. It shouldn't be right now since its the only place using `stack_pointer` besides `copy_west()`, but I don't know if the lightweight stack might be used for other tasks inside of which `unifying_equality()` might be called (maybe tail recursion?). I considered putting an `assert!(frame_pointer == stack_pointer)` at the start of this to check.
- We loop by popping a `(*mut Noun, *mut Noun)` off the stack and testing their raw equality, mug equality, etc as usual.
- When two cells are found to be unequal, their head and tail are allocated, and pointers to those allocations are added to the lightweight stack, and we loop again.

I haven't thought through how the memory looks at the end. Should we be expecting that all allocations made for this have been reclaimed? Basically I'm not sure which of these two lines is right
```
    assert!(stack.alloc_pointer == start_alloc_pointer)
    stack.alloc_pointer = start_alloc_pointer;
```

drbeefsupreme:
I rushed through `copy_pma()`, `mug()`, `jam()`, and `cue()` mostly just copying what I did for `copy_west()` and `unifying_equality()` just to get something that builds. It does build and I do some dumb things (like made `alloc_pointer` public) just to get it building, but here's the branch for it:

https://github.com/urbit/ares/tree/jon/stack-split-mem-rewrite-gogo

rust/ares/src/mem.rs
```diff
@@ -23,369 +31,391 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
+    /** Alloc pointer for the current stack frame. */
+    alloc_pointer: *mut u64,
     /** MMap which must be kept alive as long as this NockStack is */
     memory: MmapMut,
 }
 
 impl NockStack {
+    /**  Initialization
+     * The initial frame is a west frame. When the stack is initialized, a number of slots is given.
+     * We add two extra slots to store the “previous” frame pointer and allocation pointer. For the
```
ashelkovnykov, line 57: Typo: we add three extra slots

rust/ares/src/mem.rs
```diff
@@ -23,369 +31,391 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
+        // I think not - you can cast a const raw pointer as a mut raw pointer just fine
+        let frame_pointer = unsafe { start.add(top_slots + RESERVED) } as *mut u64;
+        let stack_pointer = frame_pointer;
+        let alloc_pointer = unsafe { start.add(size) } as *mut u64;
         unsafe {
-            *frame_pointer = frame_pointer.add(size) as u64;
-            *frame_pointer.add(1) = ptr::null::<u64>() as u64;
+            *frame_pointer = ptr::null::<u64>() as u64;            // "frame pointer" from "previous" frame
+            *frame_pointer.add(STACK) = ptr::null::<u64>() as u64; // "stack pointer" from "previous" frame TODO is this right?
+            *frame_pointer.add(ALLOC) = start as u64;              // "alloc pointer" from "previous" frame
```
ashelkovnykov, line 76: Shouldn't these be `*frame_pointer.sub`?

ashelkovnykov, line 76: Shouldn't this also start as `null`, since there is no previous allocation pointer?

rust/ares/src/mem.rs
```diff
@@ -23,369 +31,391 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
             frame_pointer,
+            stack_pointer,
+            alloc_pointer,
             memory,
         }
     }
 
+    /** Checks if the current stack frame has West polarity */
+    //TODO the interpreter won't know when this invariant is violated unless
+    // stack_pointer == alloc_pointer, for that we should implement a guard page
```
ashelkovnykov, line 90: > guard page

look at me - I'm the road system now

rust/ares/src/mem.rs
```diff
@@ -23,369 +31,391 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
+    #[inline]
+    fn is_east(&self) -> bool {
+        if self.stack_pointer > self.alloc_pointer {
+            true
+        } else if self.stack_pointer < self.alloc_pointer {
+            false
+        } else {
+            panic!("stack_pointer == alloc_pointer");
+        }
+    }
```
ashelkovnykov, line 112: I don't think we need this, given that there's only two polarities and we have a boolean test for West polarity. Implicitly confirmed by the fact that it's unused (though I understand that this is an unfinished draft).

rust/ares/src/mem.rs
```diff
@@ -23,369 +31,391 @@ fn indirect_raw_size(atom: IndirectAtom) -> usize {
...
+    // that the pointer is between the current allocation pointer and
+    // the previous stack pointer.
+    //
+    // on the other hand, not every address between the alloc_pointer and
+    // previous frame_pointer is a valid address for an allocated object.
+    // i believe the FP-relative slots store the addresses of the allocated
+    // objects for that frame. but they could also point to more senior
+    // memory outside of that frame. so maybe i first check that the pointer
+    // is in range, then check to see if that is a legitimate pointer or
+    // if its pointing to the middle of an allocated object
```
ashelkovnykov, line 131: My gut reaction is that an address *is* "in frame" if it's between the allocation and previous stack pointer. If that address is not pointing to an allocated object (e.g. it's pointing to the middle of an object), that's a separate error: we have a problem with our pointer arithmetic somewhere.

Do you have an example that contradicts the above interpretation?

ashelkovnykov (commented):
Just a really quick look over without diving too deeply, since this is still WIP

rust/ares/src/mem.rs
drbeefsupreme: There's some logic to what these should be set to when popping off the last frame, which I plan to discover based on what breaks at that point. For reference, the current (non split stack) implementation also has a non-null fake "previous" pointer for the `stack_pointer`.

drbeefsupreme: Yup, thanks!

rust/ares/src/mem.rs
drbeefsupreme: Yeah I wrote it for symmetry but you're right. No need to keep it around.

rust/ares/src/mem.rs
drbeefsupreme: Agreed on what "in frame" means here.

Also that's a good point that its a separate error if a pointer is not pointing to an allocated object.

drbeefsupreme:
So I think I have working implementations of `copy_east()`, `copy_west()`, `cue()`, `jam()`, `unifying_equality()`, and `mug()`. `cargo run test_data/hurray.jam` works fine. But it crashes partway through running `cargo run test_data/decrement2.jam` (which just decrements 100 once, instead of the thousand times for the other decrement jamfile).

I've been littering printfs everywhere trying to find the culprit. The `subject` changes to a cell of what appears to be two identical pointers into the memory arena, but I've been running in circles trying to figure out how this is happening.

```
...
NockStack pop 7
frame_pointer: 0x7f68fea000c0
stack_pointer: 0x7f68fea000c0
alloc_pointer: 0x7f69029ffd28
interpret loop subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 0 100]
next instruction Nock9ComputeResult
Nock9ComputeResult0 subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 0 100]
Nock9ComputeResult1 subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 0 100]
Nock9ComputeResult2 subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 1 100]
slot() noun: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 1 100]
slot loop
Ok(cell) = noun.as_cell()
slot loop
push_formula formula: [6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7]
NockStack push 8
frame_pointer: 0x7f69029ffcf0
stack_pointer: 0x7f69029ffcf0
alloc_pointer: 0x7f68fea000c0
interpret loop subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 1 100]
next instruction Nock6ComputeTest
Nock6ComputeTest subject: [[6 [5 [0 7] 4 0 6] [0 6] 9 2 [0 2] [4 0 6] 0 7] 1 100]
Nock6ComputeTest formula: [5 [0 7] 4 0 6]
push_formula formula: [5 [0 7] 4 0 6]
push_formula(): 5
NockStack push 9
frame_pointer: 0x7f68fea000f0
stack_pointer: 0x7f68fea000f0
alloc_pointer: 0x7f69029ffcf0
push_formula(): 5 *local_noun_pointers: [0 7] [4 0 6]
interpret loop subject: [140088992333040 140088992333040]
next instruction Nock5ComputeLeftChild
Nock5ComputeLeftChild subject: [140088992333040 140088992333040]
Nock5ComputeLeftChild formula: [0 7]
push_formula formula: [0 7]
NockStack push 10
frame_pointer: 0x7f69029ffcc8
stack_pointer: 0x7f69029ffcc8
alloc_pointer: 0x7f68fea000f0
0 => *stack.local_noun_pointer(1) 7
interpret loop subject: [140088992333040 140088992333040]
next instruction Nock0Axis
local_noun_pointers 0x7f69029ffce0 0x7f69029ffce8
*local_noun_pointers 4 7
Nock0Axis subject: [140088992333040 140088992333040]
slot() noun: [140088992333040 140088992333040]
slot loop
Ok(cell) = noun.as_cell()
slot loop
thread 'main' panicked at 'Axis tried to descend through atom: 140088992333040', src/interpreter.rs:612:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`140088992333040` in hex is `0x7f69029ffcf0`, which we see present above:
```
NockStack push 8
frame_pointer: 0x7f69029ffcf0
stack_pointer: 0x7f69029ffcf0
alloc_pointer: 0x7f68fea000c0
```
So, it looks like the `subject` might be turning into the tuple `[frame_pointer stack_pointer]` somehow. But putting printfs before and after every place the `subject` is changing, I don't see where its actually happening. The `interpret()` loops and we suddenly have a new subject - or so it seems.

@eamsden anything sticking out that looks strange?

drbeefsupreme:
Ok I just noticed that `0x7f69029ffcf0` is also the `alloc_pointer` in the same frame. When you `stack.push()`, the new `alloc_pointer` is set to the previous `stack_pointer`. The `subject`, while being a local variable for `interpret()`, is also stored in FP-relative slots. So I'm guessing the `subject` is being overwritten somehow when pushing a new stack frame.

Strange that it seems to get through a fair number of steps before this happens (including at least one similar looking sequence). But that should also narrow it down somewhat. That probably rules out something as simple as the `push()` or `pop()` logic being wrong, and instead something particular to the `alloc_pointer` or (more likely I think) the `stack_pointer` in one of the several functions making use of the lightweight stack.

drbeefsupreme:
I added another printf to `Nock6ComputeTest` after `push_formula` and see that the subject actually changes after running `push_formula`. The formula it pushes uses Nock 5, but there's nothing there that changes the subject. Pretty confused as to what's happening at this point.

drbeefsupreme:
So I noticed another thing this morning that may or may not be related - the `preserve()` call for the `hamt` structure is incorrect. Right now, when a frame is popped, a function called `pre_copy()` is called which moves the reserved FP-relative slots to store the pointers for the previous frame to be alloc_pointer-relative slots. At this point, no more allocations are supposed to be made to avoid overwriting them.

However, the `hamt` `preserve()` call, which always happens after `pre_copy()`, does make an allocation, for its traversal stack. This should be easy to fix - it will probably end up looking like `copy_west()`/`copy_east()` and just allocate further into the free memory adjacent to from-space, away from the slots holding the previous frame pointers.

But - it also looks like the `if in_frame()` statement for `hamt` `preserve()` never actually catches anyways - either for split-stack or non-split stack. So this probably isn't related to the above subject mysteriously changing bug.

drbeefsupreme:
The last batch of commits cleans things up a bit.

I think the main trouble that's left is surrounding the behavior preparing to pop a stack frame. When we evacuate allocations in the current frame to the previous frame, we destroy the FP-relative slots and lightweight stack. To save the pointers to the previous frame, we call `pre_copy()`, which copies them into the free space adjacent to the allocation arena of the current frame.

Then, in order to do noun traversal for e.g `copy_west()`, we implement another lightweight stack in the free memory adjacent to those pointers. Then despite the current frame being a e.g. west frame, this lightweight stack is oriented like an normal east frame's lightweight stack. Furthermore, finding the saved pointers requires calling a different set of functions.

This makes things pretty confusing for the programmer. They have to know whether they're working before or after `pre_copy()` has been called in order to know which functions to call. I think it will be more ergonomic for `NockStack` to instead track whether `pre_copy()` has been called yet on the current frame, and have one unified interface for working with the saved pointers and the lightweight stack. So my next batch (probably later today) will be an attempt at implementing this.

drbeefsupreme:
Ready for review. A few things I wish to draw attention to:

1) `hamt::preserve()` uses a lightweight stack slightly differently than all the rest. It might make sense to change it to use to the lightweight stack interface, but I do not even know what `hamt` is for exactly just yet.
2) There is now a flag for `NockStack` called `pc` that tracks whether or not `pre_copy()` has been called on the current stack frame. This definitely improves ergonomics, but the other choice that I think there is would be to use an `enum` to say whether the lightweight stack is east-facing or west-facing. I decided to go with the boolean since this created a naming catastrophe with `_west` and `_east` functions.
3) You have to know what is on the lightweight stack to pop it - i.e. you called `stack_pop::<foo>()` instead of just `stack_pop()`. This feels wrong but I don't know a way around it.

rust/ares/src/main.rs
```diff
@@ -43,7 +43,11 @@ fn main() -> io::Result<()> {
         let in_map = Mmap::map(&f)?;
         let word_len = (in_len + 7) >> 3;
         let (mut atom, dest) = IndirectAtom::new_raw_mut(&mut stack, word_len as usize);
-        write_bytes(dest.add(word_len as usize - 1), 0, 8);
+
+        //write_bytes(dest.add(word_len as usize - 1), 0, 8);
+        //TODO I'm not sure what this +7 is really supposed to be.
+        write_bytes(dest.sub(word_len as usize + 7), 0, 8);
```
drbeefsupreme, line 49: I also don't know what is going on here. I found `+7` through trial and error.

rust/ares/src/hamt.rs
```diff
@@ -444,7 +444,8 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
             let dest_buffer = stack.struct_alloc_in_previous_frame(self.0.size());
             copy_nonoverlapping(self.0.buffer, dest_buffer, self.0.size());
             self.0.buffer = dest_buffer;
-            let traversal_stack = stack.struct_alloc::<(Stem<T>, u32)>(6);
+            //TODO Am unsure if I should be using the lightweight stack here instead
```
eamsden, line 447: no, copying cannot use the lightweight stack as its in the way of the copy

eamsden (changes requested):
really need to know what's going on with the magic # change, otherwise looking good

rust/ares/src/main.rs
drbeefsupreme: I was hoping you could tell me! The commented out line is what was there pre-split stack. When I first got it compiling, this was one of the first lines it broke at, so I fiddled with the magic number until it worked and moved on, intending to come back to it later. I figured if you could tell me what the commented line was doing, I could figure out why I had to change the magic number (I think its related to allocations going at high memory at the start instead of low memory, but I don't know what its doing either way).

rust/ares/src/hamt.rs
drbeefsupreme: So once we start copying, we actually do still have a lightweight stack, it just switches polarities. So for a west frame, the lightweight stack is adjacent to the frame pointer on the west side. Once we go into copying mode (`pre_copy()` is called so `NockStack.pc == true`), the lightweight stack moves into the free space east of allocations. `copy_west()` and `copy_east()` use the lightweight stack this way to traverse nouns to be copied.

So that's what I meant here - to use the "copying mode lightweight stack". Since `hamt::preserve()` is called when `pc==true`, we are no longer allowed to allocate since we have moved the saved pointer for the previous frame adjacent to the allocation arena for that frame. But there is still more free space next to that. So this is where e.g. `copy_west()` uses a lightweight stack. For `hamt::preserve()`, we instead have

```
let traversal_stack = stack.free_alloc::<(Stem<T>, u32)>(6);
```
which called `free_alloc()`, which is allocating in this free space that would ordinarily be occupied by the lightweight stack in the copying phase. This itself is also a stack, so my comment meant to imply that I should be using the lightweight stack interface here (`stack_push()`, `stack_pop()`, etc) instead, which knows whether or not we're in the copying phase or not, so it would be running in the same space that the `copy_west/east()` lightweight stack does.

drbeefsupreme:
Having done enough of #66 to know how the split stack interface is going to be used for it, at this point the only thing that's gone on the lightweight stack are nouns. If this is always the case, then the weirdness I'm worried about with respect to needing to know the type of what's on top of the stack in order to pop it wouldn't actually be a problem.

drbeefsupreme:
So the choice of where the `frame_pointer` and what the saved pointers are isn't obvious, but I wrote up my rationale here:

https://github.com/urbit/ares/pull/64#discussion_r1276404747

rust/ares/src/main.rs
ashelkovnykov: `dest` is a pointer to the data buffer for the indirect atom, so you definitely don't want `sub`, as that's going to redirect to the mug or size fields (at best).

`dest.add(word_len as usize - 1)` is going to result in a pointer to the final word of memory for the data buffer of the indirect atom meant to hold the jammed noun. Therefore, the entire `write_bytes` call is just overwriting the final word of memory for the indirect noun data buffer with `0`s. This is so that after the `copy_nonoverlapping` call, the trailing bytes of the indirect atom buffer (the ones beyond `size`) are all `0`. I'm not sure what the exact reason for this is, but I can think of a couple potential reasons:
- consistent mugs from reading all words of memory in the buffer (not just size)
- no garbage data in the "extra" bytes
- easy invariant to check that indirect atoms aren't corrupt
- errors in loading indirect atoms caught quickly by the following call to `normalize`

TL;DR - You want to keep this line as-is. If the code is failing at this line, it has some other issue that needs to be resolved.

ashelkovnykov (changes requested):
Partial review. Will finish the rest shortly.

rust/ares/src/mem.rs
```diff
@@ -412,15 +383,15 @@ impl NockStack {
...
                             {
                                 match allocated.as_either() {
                                     Either::Left(mut indirect) => {
                                         // Make space for the atom
-                                        let new_indirect_alloc = other_stack_pointer;
-                                        other_stack_pointer =
-                                            other_stack_pointer.add(indirect_raw_size(indirect));
+                                        other_alloc_pointer =
+                                            other_alloc_pointer.sub(indirect_raw_size(indirect));
+                                        let new_indirect_alloc = other_alloc_pointer;
```
ashelkovnykov, line 394: If we replace explicit usage of `self.prev_alloc_pointer_pointer()` via `other_alloc_pointer` with calls to `struct_alloc_in_previous_frame`, we can encapsulate all allocation to a senior frame in one function, unify `copy_east` & `copy_west` into a single function, and remove the manual reconfiguration of the senior frame's allocation pointer (lines `443` and `540`).

ashelkovnykov (changes requested):
Lots of opportunities to unify some functions. Should be very sleek & elegant indeed, afterwards.

rust/ares/src/mem.rs
```diff
@@ -113,291 +106,269 @@ impl NockStack {
...
-    pub fn in_frame<T>(&self, ptr: *const T) -> bool {
+    pub unsafe fn in_frame<T>(&self, ptr: *const T) -> bool {
         let ptr_u64 = ptr as *const u64;
-        match &self.polarity {
-            Polarity::East => ptr_u64 >= self.stack_pointer && ptr_u64 < self.frame_pointer,
-            Polarity::West => ptr_u64 >= self.frame_pointer && ptr_u64 < self.stack_pointer,
+        let prev = *self.prev_stack_pointer_pointer();
+        if self.is_west() {
+            ptr_u64 >= self.alloc_pointer
+                && ptr_u64 < prev
```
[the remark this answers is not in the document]
eamsden: this is more correct

drbeefsupreme:
@ashelkovnykov Thanks for your comments, I'll get to them in the morning!

rust/ares/src/hamt.rs
drbeefsupreme: This is a question about the right way to traverse the HAMT to copy it into the allocation arena of the previous frame - I don't think where the HAMT is being copied to (the previous frame's allocation arena) is under dispute.

It's really just a question about what is the least confusing interface. After mulling it over for the last week, I'm pretty sure that I should be using the lightweight stack here. I'll try writing it that way and then we can just compare and see which satisfies the principle of least surprise.

rust/ares/src/interpreter.rs
```diff
@@ -87,6 +87,7 @@ pub fn interpret(
         loop {
             match noun_to_work(*(stack.local_noun_pointer(0))) {
                 Done => {
+                    stack.pre_copy();
```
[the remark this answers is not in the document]
drbeefsupreme: This is a great idea. I'm going to think a little bit about the details but I'll do something along these lines. I think the goal with `pre_copy()` and everything related to it should be to make it so the programmer doesn't even need to know that it exists unless absolutely necessary.

rust/ares/src/mem.rs
```diff
@@ -660,88 +614,215 @@ impl NockStack {
...
+     * frame situation is the same, swapping west for east.
+     *
+     * When pc == true, the lightweight stack is used for copying from the current
+     * frame's allocation arena to the previous frames. */
+
+    /** Push onto the lightweight stack, moving the stack_pointer. Note that
+     * this violates the _east/_west naming convention somewhat, since e.g.
+     * a west frame when pc == false has a west-oriented lightweight stack,
+     * but when pc == true it becomes east-oriented.*/
+    pub unsafe fn stack_push<T>(&mut self) -> *mut T {
```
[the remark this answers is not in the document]
drbeefsupreme: Agreed

rust/ares/src/mug.rs
```diff
@@ -170,6 +167,7 @@ pub fn mug_u32(stack: &mut NockStack, noun: Noun) -> u32 {
         }
     }
     unsafe {
+        stack.pre_copy();
```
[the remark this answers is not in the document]
drbeefsupreme: Yeah we actually don't need a new stack frame at all here. I removed it.

rust/ares/src/mem.rs
```diff
@@ -582,74 +536,74 @@ impl NockStack {
...
-     */
     pub unsafe fn copy_pma(&mut self, noun: &mut Noun) {
-        assert!(self.polarity == Polarity::West);
-        let work_start = self.stack_pointer;
-        self.stack_pointer = self.stack_pointer.add(2);
-        *(self.stack_pointer.sub(2) as *mut Noun) = *noun;
-        *(self.stack_pointer.sub(1) as *mut *mut Noun) = noun as *mut Noun;
+        assert!(self.is_west());
+        let noun_ptr = noun as *mut Noun;
+        self.stack_pointer = self.alloc_pointer.sub(RESERVED + 1);
```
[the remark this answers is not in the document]
drbeefsupreme: Good catch.

rust/ares/src/mem.rs
```diff
@@ -660,88 +614,215 @@ impl NockStack {
...
-            Polarity::East => self.pop_east(),
-            Polarity::West => self.pop_west(),
+        if !self.stack_is_empty() {
+            //TODO Should this actually be a panic?
+            eprintln!("Popping stack frame when lightweight stack is not empty!");
+        }
+        if self.is_west() {
+            self.pop_west()
+        } else {
+            self.pop_east()
```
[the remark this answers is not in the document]
drbeefsupreme: So this will segfault because `free_slot` calls `is_west`, but this variation does work, so I'll just do this.

```
    pub unsafe fn frame_pop(&mut self) {
        let prev_frame_ptr = *self.free_slot(FRAME) as *mut u64;
        let prev_stack_ptr = *self.free_slot(STACK) as *mut u64;
        let prev_alloc_ptr = *self.free_slot(ALLOC) as *mut u64;

        self.frame_pointer = prev_frame_ptr;
        self.stack_pointer = prev_stack_ptr;
        self.alloc_pointer = prev_alloc_ptr;

        self.pc = false;
    }
```

rust/ares/src/mem.rs
```diff
@@ -660,88 +614,215 @@ impl NockStack {
...
-            match &self.polarity {
-                Polarity::East => self.push_east(num_locals),
-                Polarity::West => self.push_west(num_locals),
+            if self.is_west() {
+                self.push_west(num_locals)
+            } else {
+                self.push_east(num_locals)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
drbeefsupreme: Required a tweak but this works.

I did notice something weird about the use of `unsafe` though. Before split stack, `pop` is an unsafe function, but `push` is a safe function with an unsafe block in it. I kept that convention here since I don't know how you decide when to declare a function unsafe and when to use an unsafe block in general.

@eamsden

rust/ares/src/mem.rs
drbeefsupreme: I figured something like this was possible. I'll definitely be implementing it. Thanks!

drbeefsupreme: https://github.com/urbit/ares/pull/60/commits/1cb3ec59604b372b35d8dc750bfdaf6ffb4172fa

rust/ares/src/interpreter.rs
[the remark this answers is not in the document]
drbeefsupreme: For some reason I thought this would be trickier than it turned out to be.

https://github.com/urbit/ares/pull/60/commits/838450da7c367619c7a02473bbc200328ead6887

rust/ares/src/main.rs
drbeefsupreme: I slept on it and am pretty sure I'm right so I went ahead and committed it.

https://github.com/urbit/ares/pull/60/commits/c34fdc65849ebe7301e493b0dcad4a67460aa0f4

rust/ares/src/mem.rs
```diff
@@ -75,36 +40,64 @@ pub struct NockStack {
...
+     * We add three extra slots to store the “previous” frame, stack, and allocation pointer. For the
+     * initial frame, the previous allocation pointer is set to the beginning (low boundary) of the
+     * arena, the previous frame pointer is set to NULL, and the previous stack pointer is set to XX */
+
     /** Size is in 64 bit words.
      * top_slots is how many slots to allocate to the top stack frame.
      */
     pub fn new(size: usize, top_slots: usize) -> NockStack {
+        //TODO memory doesn't have to be mut as written, but it was before split stack. not sure what
+        // it was accomplishing.
```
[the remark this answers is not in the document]
drbeefsupreme: Ok I took out the `mut` keyword.

rust/ares/src/hamt.rs
drbeefsupreme: https://github.com/urbit/ares/pull/60/commits/bf7fe776a0c93b1bfa8e74e8acb3fdd5839f731b

drbeefsupreme: I don't seem to actually have a jamfile ready to test `hamt::preserve()`. I'm not sure how one is ever even made, so I don't know what I'd need to write to get a jamfile that does use it. So this code is untested.

rust/ares/src/mem.rs
```diff
@@ -113,291 +106,269 @@ impl NockStack {
...
-    pub unsafe fn struct_alloc_in_previous_frame<T>(&mut self, count: usize) -> *mut T {
-        match &self.polarity {
-            Polarity::East => self.struct_alloc_in_previous_frame_east(count),
-            Polarity::West => self.struct_alloc_in_previous_frame_west(count),
+    /** Pointer to where the previous stack pointer is saved in a frame */
+    unsafe fn prev_stack_pointer_pointer(&self) -> *mut *mut u64 {
+        if self.pc == false {
+            self.slot_pointer(STACK) as *mut *mut u64
+        } else {
+            self.free_slot(STACK) as *mut *mut u64
```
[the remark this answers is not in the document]
drbeefsupreme: So I've ended up making more use of this (and analogues for the frame and stack pointer) since the first version of the PR. I tried changing it to return e.g. `*self.slot_pointer(STACK) as *mut u64` but, at least written like that, it doesn't seem you can change the pointer that way. Meaning
```
self.prev_stack_pointer() = (self.prev_stack_pointer()).add(1)
```
doesn't compile. I know there's more to the Rust raw pointer interface that I don't know about, so maybe some other kind of syntax would work. But at the moment my impression is that we're stuck with pointer pointers.

drbeefsupreme:
I believe I have addressed all comments. Would you please give it another look over @ashelkovnykov @eamsden

rust/ares/src/mem.rs
```diff
@@ -384,178 +389,81 @@ impl NockStack {
...
+                            if self.is_in_frame(allocated.to_raw_pointer())
                             {
                                 match allocated.as_either() {
                                     Either::Left(mut indirect) => {
                                         // Make space for the atom
-                                        other_alloc_pointer =
-                                            other_alloc_pointer.sub(indirect_raw_size(indirect));
-                                        let new_indirect_alloc = other_alloc_pointer;
+                                        let alloc =
+                                            self.indirect_alloc_in_previous_frame(indirect.size());
```
ashelkovnykov, line 398: Nothing wrong here, I'm just not sure of the value of `indirect_alloc_in_previous_frame` over:
```
let alloc = self.struct_alloc_in_previous_frame::<u64>(indirect.size() + 2);
```
since it's not used anywhere else.

ashelkovnykov (commented):
LGTM, pending:
- Fix to pass CI
- Reformat w/ `cargo fmt`
- Squash commits

rust/ares/src/mem.rs
drbeefsupreme: Having both `indirect_alloc_in_previous_frame()` and `struct_alloc_in_previous_frame()` is consistent with there being an `indirect_alloc()` and `struct_alloc`, and notably there is also only one use of `indirect_alloc()`. So I don't think we should remove one without removing the other to keep the interface consistent.

I'm in favor of keeping them since otherwise the programmer needs to understand the memory layout of indirect atoms in order to know that the `+ 2` is universal for all indirect atoms and not just some number that only makes sense for that one particular callsite.

drbeefsupreme:
What's the deal with squashing commits? I've only done it sparingly, such as when I'm really confused and need to test out a lot of different paths and stashing isn't enough so I use WIP commits. Many (most?) of the commits here have commit messages explaining their rationale. Squashing them would create a gigantic confusing commit message and thus make the commit history a lot less useful IMO.

If that's just the style for the Ares repo I'll do it, but I think it is a loss.

drbeefsupreme:
Not sure how to deal with this from CI:

https://github.com/urbit/ares/actions/runs/5790048220/job/15692346635

I definitely don't think I should rewrite this using `match` and the `Ordering` enum. I think its more clear as written.

ashelkovnykov:
@drbeefsupreme 

> If that's just the style for the Ares repo I'll do it, but I think it is a loss.

I don't actually know what the style is regarding squashing commits for Ares - we should confirm with @eamsden.

> Unless you mean I should go through all the commits and figure out which ones are appropriate to squash together. Which sounds tiresome, but I'd be fine with doing that.

I meant squash them all down into one commit, to be courteous of your time, but I *do* personally prefer the practice of every PR in a commit being a logical, (almost) self-contained grouping of changes.

> Many (most?) of the commits here have commit messages explaining their rationale.

This is true; I just took a look at them and they're a lot more descriptive than the usual fare one would expect from a PR with nearly 100 commits.

> Squashing them would create a gigantic confusing commit message and thus make the commit history a lot less useful IMO.

I disagree. We care about the changes and we care about the discussion around the changes, but we don't care about intermediate work. Any important discoveries or notes made during the process should be documented somewhere far more discoverable than the commit history. 87 commits is massively excessive for what comes down to ~1000 modified lines. If I want to browse a history of the changes, I don't want to have to scroll past 87 commits that are all introducing the same one feature. If I'm doing a git-bisect, I don't want to waste hops because 87 commits are all introducing one feature.

Regardless, the most important thing here is the changes themselves, so I'll go with whatever decision @eamsden makes regarding git + GitHub etiquette.

ashelkovnykov:
> Not sure how to deal with this from CI:
> 
> https://github.com/urbit/ares/actions/runs/5790048220/job/15692346635
> 
> I definitely don't think I should rewrite this using `match` and the `Ordering` enum. I think its more clear as written.

@drbeefsupreme I think our options are:
1. Do what the linter says
2. Suppress the warning globally
3. Suppress this particular warning by adding this right above the `if` statement: `#[allow(clippy::comparison_chain)]`

eamsden:
@drbeefsupreme @ashelkovnykov 

On squashing: you can always edit the commit messages. Squash together into one or 2 commits of working code

On CI: that particular linter bit is fine to suppress, ifs with comparators are good and fine

drbeefsupreme:
I was just hiding behind laziness on making more than one squash commit. I don't actually know how to do anything more complicated than that yet but I'll figure it out and squash this into a few commits, and suppress the warning.

drbeefsupreme:
Ok I squashed all the commits and kept the useful sounding commit messages, but it doesn't look like there's a way to see what code they're referring to? Is there any way to maintain that or is it just lost unless I single out the commits with useful messages and squash everything in between them?

Sorry this process is confusing for me - what the right convention is and what the results will look like when I try a new process is unclear to me.

eamsden:
> Sorry this process is confusing for me - what the right convention is and what the results will look like when I try a new process is unclear to me.

In urbit we basically do commit-per-feature, so once commit saying this introduces the split-stack is fine

eamsden (approved):

ashelkovnykov (approved):

PR #61, opened 2023-06-22
docs: add noun rep svg

drbeefsupreme:


drbeefsupreme:
One bit of confusion left: it looks like pointers are always left-shifted by 3. So I suppose that means for an indirect atom, I should mark the 3rd bit as reserved instead of a pointer bit?

eamsden (approved):
LGTM

ashelkovnykov:
@drbeefsupreme Petition for either "light green" or "yellow" to be replaced by another colour - my feeble [deuteranomalous vision](https://i.pinimg.com/originals/d4/09/96/d409961f245a62b4e97162e09cca3055.jpg) can barely tell them apart.

PR #63, opened 2023-07-07
Unit tests for the PMA

ashelkovnykov:
**Summary:**
Unit tests for the PMA C module

**Unimplemented tests:**
In conversation on July 19-20, 2023, the decision to remove the PHK aspects of the PMA from the design was reconfirmed. This means that there are now two upcoming refactors to the PMA: the switch from an Array-based page directory to a B+ Tree page directory, and the switch to pages as the smallest unit of persistent allocation. As such, unit tests for the following functions were skipped:
- `_pma_malloc_bytes`
- `_pma_write_page_offset`
- `_pma_write_page_status`
- `_pma_sync_dirty_pages`

The following logic, which would otherwise be tested in `pma_load`, is not:
- Syncing unsynced changes to the page directory upon load after a crash
- Correctly computing the free page and free page run caches
    - Correctly not joining two consecutive free pages into a free page run unless they're also adjacent on disk (not just in memory)

Tests for other functions which are planned to be removed in the upcoming refactors were already completed by the time the aforementioned decision was made.

In addition, the following logic, which would otherwise be tested by integration tests, is not:
- Wrapping the dpage cache around (`head` becomes lower than `tail`)
- Completely filling up the dpage cache
- Completely filling up the dirty page cache

**Calling tests from `cargo`:**
Currently, the unit and sanity tests use the same virtual memory addresses. Since Rust tests run in parallel by default, this was causing the tests to clobber each other and fail. Rust has no notion of ordered tests, or test dependencies. In fact, it doesn't even have the notion of disabling parallel tests: instead, this is accomplished by limiting the number of threads that `cargo test` is allowed to use to `1`.

To get the tests to pass, I've "disabled parallel tests". Alternative solutions which would allow the tests to not clobber each other are:
1. Go through the unit tests and change them to use different virtual memory addresses
2. Move the unit and sanity tests into separate crates

I think solution 1 is possible, but I'm not certain. It wouldn't take very long to confirm whether it is, and make the required changes if so. I can do this now, rather than later if anyone insists on getting it done in this particular PR.

Solution 2 is stupid, annoying, and (shockingly) the conventional Rust solution.

**Unit test warnings:**
There are only two types of warnings still present in the unit tests: warnings about casts between integer and pointer types, and warnings about comparisons between integer and pointer types. There are a lot of both.

The cast warnings can be disabled with a `-Wno-int-conversion` flag to gcc. Unfortunately, the comparison warnings are at the level of the language standard itself, meaning the only way I could find to turn them off was to turn off *all* warnings with the `-w` gcc flag. Obviously, this is not ideal: there are certain warnings in the unit tests about which we'd prefer to hear.

### DIFF
--- a/.github/workflows/ares-shared.yml
+++ b/.github/workflows/ares-shared.yml
@@ -39,4 +39,4 @@ jobs:
         run: cargo build --release --verbose
 
       - name: Run tests
-        run: cargo test --verbose
+        run: cargo test --verbose -- --test-threads=1
--- a/rust/ares/build.rs
+++ b/rust/ares/build.rs
@@ -13,6 +13,14 @@ fn main() {
             std::process::exit(-1);
         }
     }
+
+    cc::Build::new()
+        .file("./src/pma/test/malloc.c")
+        .opt_level(0)
+        .flag("-g3")
+        .flag("-Wno-int-conversion")
+        .flag("-w")
+        .compile("test_pma_malloc_unit");
 }
 
 fn debug() {
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -1,6 +1,6 @@
 use crate::assert_acyclic;
 use crate::noun::{Atom, Cell, CellMemory, IndirectAtom, Noun, NounAllocator};
-use crate::snapshot::pma::{pma_in_arena, pma_malloc};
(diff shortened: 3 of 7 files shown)


eamsden (approved):

eamsden:
@ashelkovnykov please squash before merging

barter-simsum (approved):

PR #64, opened 2023-07-24
docs: split stack

drbeefsupreme:
Swaps out `stack.md` for `split-stack.md` and `noun-rep.md`


docs/split-stack.md
```diff
@@ -0,0 +1,224 @@
...
+These slots are sometimes called "frame pointer relative slots", or FP-relative
+slots. For a west stack frame, these slots are immediately west of the frame
+pointer, with the first three slots west of the frame pointer being reserved to
+store the pointers of the previous frame. These slots extend to the previous
+allocation pointer.
+
+We emphasize that the frame pointer does _not_ mark a boundary for the stack
+frame. For a west frame, the pointers are always ordered as previous allocation
+pointer is less than the current frame pointer is less than or equal to the
+current stack pointer. For an east frame, substitute "less" with "greater". The
```
ashelkovnykov, line 90: >  For a west frame, the pointers are always ordered as previous allocation
pointer is less than the current frame pointer is less than or equal to the
current stack pointer.

This is a really confusing sentence. It would probably be clearer as a bullet point list. Ex:

For a west frame:
- `prev_alloc_ptr = (frame_ptr - 3)`
- `stack_ptr >= frame_ptr`
- `*(frame_ptr - 2) = prev_alloc_ptr`
- `*(frame_ptr - 1) = prev_stack_ptr`
- `*(frame_ptr) = prev_frame_ptr`

Note that this also includes information from the "Execution Phase" section as well.

While writing this, it occurred to me: why do we need a slot for the previous allocation pointer, if the contents are always `frame_ptr +/- 2`, depending on the polarity? I recall this came up early during the work, but I don't remember what the reason was.

docs/split-stack.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Execution phase
+
+Three pointers are kept in registers - the frame pointer, stack pointer, and
+allocation pointer. They are arranged for west and east frames as follows:
+
+West frame (execution phase):
+|----aXXXFLLSoooAYYYs----|
+East frame (execution phase):
+|----sYYYAoooSLLFXXXa----|
```
ashelkovnykov, line 42: I think this diagram does more to confuse than to enlighten. My two big issues with it are:
1. It doesn't distinguish between the FP-relative slots and stack slots
2. It makes it appear as if allocation, stack, and frame pointers take up space in memory, as opposed to pointing to parts of memory

I think we'd be better served by having diagrams of the kind @drbeefsupreme made for describing nouns, and then using these notes to refer to them and explain the *why* of the design, not the what.

ashelkovnykov (changes requested):
Partial review: made it to the "Stack Pointer" section of `split-stack.md` and had some strong issues with the documentation that I think might affect the remainder of it, so I ceased review.

Changes to the other files LGTM, though.

docs/split-stack.md
drbeefsupreme: 1. It does - that's what `X` and `L` are. But I still agree that the diagram is hard to read.
2. This is a good point

I have this in an .svg diagram already, but I'm still tweaking it since I want to use the same one to describe an example of lightweight stack usage and I keep running into issues that make it hard to read. I'll give it another go sometime this week or early next week.

docs/split-stack.md
drbeefsupreme: Agreed that a list of equations is more lucid.

> While writing this, it occurred to me: why do we need a slot for the previous allocation pointer, if the contents are always frame_ptr +/- 2, depending on the polarity? I recall this came up early during the work, but I don't remember what the reason was.

That's only true if there are no additional FP-slots beyond the reserved slots. Right now, it's actually

- prev_alloc_ptr = (frame_ptr - 3 - slots)

You could do this though - you could either put the saved pointers west of `frame_pointer` and additional slots east of `frame_pointer, or put the frame pointer at the westmost word of a west frame, and FP-relative slots to the east. You'd need to save the initial value of `stack_pointer` to tell when it is empty. So you no longer need to include the previous allocation pointer as you point out, but now need to include the initial value of `stack_pointer` for that frame to tell when the lightweight stack is empty.

So you use the same amount of memory, but you end up with a different interface for the programmer. I think the way it currently works ultimately has a better interface. One reason for this is as follows.

When a frame is in "cleanup phase", you have to save the parent frame's pointers in the free space adjacent to the current frame's `alloc_pointer`. Then when you start copying from the current frame's allocations to the previous frame's, you adjust the value of  `prev_alloc_pointer` as you go. So `prev_alloc_pointer` can no longer be calculated relative to the current frame's `frame_pointer`. Thus there's logical continuity going from

```(prev_alloc_pointer, prev_frame_pointer, prev_stack_pointer)` -> `(prev_alloc_pointer, prev_frame_pointer, prev_stack_pointer)```
when you go from execution to cleanup phase, vs

```(prev_frame_pointer, prev_stack_pointer, initial_stack_pointer)` -> `(prev_alloc_pointer, prev_frame_pointer, prev_stack_pointer)```
which is what you get when you have the initial value of `stack_pointer` as a saved pointer instead of `prev_alloc_pointer`.

I thought I had a second reason but I forgot by the time I got here. Just the one reason should be enough though.

Comment (ashelkovnykov) [reply docs/split-stack.md:90]: That makes sense - I was misunderstanding how slots were tracked. Thank you for the explanation.

**Comment by ashelkovnykov** [comment]
@drbeefsupreme Thanks for your response. I'll continue reviewing from where I left off after I review #60 - I think It'll be easier to gauge the documentation after I understand the exact behaviour of the code.

**Comment by drbeefsupreme** [comment]
I'll add in some edited version of my response to make things more clear.

File: docs/split-stack.md [hunk 3]
```diff
@@ -0,0 +1,224 @@
...
+The new stack pointer is set to be equal to the new frame pointer.
+
+The new allocation pointer is set to be equal to the previous frame's stack pointer.
+
+The relative position of the stack and allocation pointers determines whether a
+stack frame is called west or east. If the stack pointer is west of the 
+allocation pointer, the polarity is west. If the stack pointer is east of the
+allocation pointer, the polarity is east. Thus, by the process outlined about,
+we see that when the current stack frame has a west polarity, the following one
+will always have an east polarity, and vice versa.
```
ashelkovnykov, line 144: Since the frame pointer never moves and the allocation pointer always moves towards it, I think it would be better to explain polarity with reference to the frame and allocation pointers instead of the stack and allocation pointers.

eamsden: This is a good point. @drbeefsupreme can you make this change when you get a chance please? Since we merged the split stack I'd like to get the docs updated ASAP

drbeefsupreme: https://github.com/urbit/ares/pull/64/commits/0e3245bfd70f2940d33b9339408ca845cd241532

docs/split-stack.md
drbeefsupreme: Here's an svg: https://github.com/urbit/ares/pull/64/commits/bdf48555292b9a39ca6c9c9f85dd29c91fa491b9

drbeefsupreme:
Some new revisions, including a much-needed diagram. Can you take another look @ashelkovnykov?

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+allocation pointer (inclusive) and west of the current stack pointer
+(non-inclusive). The "heap side" for a west frame is the memory eastward of the
+allocation pointer (inclusive) and westward of the previous stack pointer
+(non-inclusive).
+
+Similarly, the stack side for an east frame is the memory westward of the
+previous allocation pointer (non-inclusive) and eastward of the current stack
+pointer (inclusive). The heap side for an east frame is the memory westward of
+the current allocation pointer (non-inclusive) and eastward of the previous
+stack pointer (inclusive).
```
ashelkovnykov, line 115: If my above comment is applied, the concepts of "stack side" and "heap side" have already been explained. I also feel that this part is a little unclear, in that the slots and frame pointers are lumped in with the stack. If this section is kept at all, it might be clearer to have something like the following:
```
The stack for a west frame is the memory eastward of the frame pointer
(inclusive) and westward of the current stack pointer (exclusive). The heap 
or a west frame is the memory eastward of the current allocation pointer
(inclusive) and westward of the previous stack pointer (exclusive).

Similarly, the stack for an east frame is the memory westward of the frame
pointer (exclusive) and eastward of the current stack pointer (inclusive).
The heap for an east frame is the memory westward of the current allocation 
pointer (exclusive) and eastward of the previous stack pointer (inclusive). 

The slots and registers for a frame are always stored in the memory between
the previous allocation pointer and the frame pointer (inclusive).
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+
+#### Frame pointer
+
+The frame pointer is the only pointer of the three which never moves.
+
+Each stack frame has a number of slots (each one machine word in length) in the
+memory arena. There are always at least three slots. These three slots are
+reserved to store the pointers for the previous frame. Further slots may be used
+to store things like the return of a function or the Nock instruction being run in
+that stack frame.
```
ashelkovnykov, line 127: Above, the three reserved slots are referred to as "registers"

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+
+We emphasize that, unlike many common stack schemas, the frame pointer does
+_not_ mark a boundary for the stack frame. For a west frame, the memory west of
+`frame_ptr` are slots in which other pointers are stored. If the lightweight
+stack is non-empty, then the memory east of `frame_ptr` contains elements of the
+lightweight stack.
+
+The ultimate reason for this is that it improves the ergonomics of the pointer
+interface. A deeper explanation is given [at the end](#frame-boundary), since
+you need to understand what all the pointers are for and what the cleanup phase
```
ashelkovnykov, line 143: "... you need to **first** understand ... "

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+number of slots in machine words.
+
+The new stack pointer is set to be equal to the new frame pointer.
+
+The new allocation pointer is set to be equal to the parent frame's stack pointer.
+
+The relative position of the frame and allocation pointers determines whether a
+stack frame is called west or east. If the frame pointer is west of the 
+allocation pointer, the polarity is west. If the frame pointer is east of the
+allocation pointer, the polarity is east. Thus, by the process outlined about,
```
ashelkovnykov, line 194: "... the process outlined **above** ..."

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+A current allocation pointer
+a previous allocation pointer
+F current frame pointer
+f previous frame pointer
+S current stack pointer
+s previous stack pointer
+X Current memory slots
+L Current lightweight stack
+Y Current allocations
+o free memory
```
ashelkovnykov, line 242: See above: the SVG is so much better that this ASCII diagram should be axed and the reader directed to the visual reference

ashelkovnykov (changes requested):

docs/split-stack.md
eamsden: Agreed. If the SVG is committed it could even be included as a markdown image reference:

```markdown
![alt text](relative_url.svg)
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+Lower memory is directionally "west", and higher memory is directionally "east".
+A frame whose "stack side" (to be defined shortly) is lower in memory than their
+"heap side" is called a "west frame". On the other hand, a frame whose stack
+side is higher in memory than their heap side is called an "east frame".
+
+Computational frames are always pushed in an alternating west/east fashion (with
+the first frame always being a west frame).
+Thus the two stacks logically form one single computational stack.
+The stack on which the stack side of the current frame is located is termed the "polarity".
+Thus, the current stack frame may be said to have "west" or "east" polarity.
```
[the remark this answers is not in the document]
eamsden: :+1:

docs/split-stack.md
eamsden: there's no registers in the stack. There are "registers" (fields of the NockStack struct) which define the current state of the stack, but words referenced by a fixed offset from the frame pointer are called "slots"

docs/split-stack.md
eamsden: The only thing which is correct to call a "register" in this document is the *current* stack/frame/allocation pointers.

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+L Current lightweight stack
+Y Current allocations
+o free memory
+
+#### West and East frames
+
+For a west frame with `n` slots, the pointers are located in memory as follows:
+
+ - `frame_ptr <= stack_ptr < alloc_ptr`
+ - `prev_alloc_ptr = frame_ptr - 3 - n`
```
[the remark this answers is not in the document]
ashelkovnykov: Scratch that - forgot that we're talking about the pointer, not the value *at* the pointer

docs/split-stack.md
ashelkovnykov: Updated comment to remove calling the three permanent slots "registers"

docs/split-stack.md
ashelkovnykov: Or an `img` tag so that we can control the width/height of the image, plus link to the original

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+ - Pushing the lightweight stack moves `stack_ptr` eastward.
+
+#### Copying
+
+After the frame has moved into the cleanup phase, copying from the current
+allocation arena to the previous one can begin.
+
+When popping a frame, the result register is examined to provide a collection
+root. Pointers which fall within the current frame's allocation arena are
+recursively chased and copied to the parent frame's allocation arena. This is
```
ashelkovnykov, line 280: See above.
```
Pointers to result components which fall within the current frame are tracked down recursively from the root and copied to the parent frame's stack.
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+head and tail pointers are pushed onto the lightweight stack. Then when we pop, say, the
+head, we check to see if its located within the current stack frame or a senior
+one. If it is the current stack frame we copy it to the previous stack frame,
+otherwise it is left alone.
+
+This ensures that data allocated in this frame which is live from the returned
+result is not lost when the frame is popped.
+
+The reason that this works is due to the following invariant: junior memory may
+only ever point to other memory within in the same stack frame, or more senior
```
ashelkovnykov, line 291:
```suggestion
only ever point to other memory within the same stack frame, or more senior
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+memory. Senior memory may never point into junior memory.
+
+Copying a cell results in a new cell allocation to the parent frame's allocation
+arena, and pushing both constituent noun pointers onto the lightweight stack with the
```
ashelkovnykov, line 295:
```suggestion
Copying a cell results in a new cell allocation to the parent frame's
stack, and pushing both constituent noun pointers onto the lightweight stack with the
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+ - `pop<T>()` - pop a `T` off the lightweight stack.
+ - `top<T>() -> *mut T` - peek at a `T` on the top of the lightweight stack.
+ - `stack_is_empty() -> bool` - true when lightweight stack is empty.
+
+### Why `frame_ptr` does not mark a frame boundary {#frame-boundary}
+
+Let's instead consider what would happen if `frame_ptr` for a west frame were
+the westmost word of the frame - i.e. it marks a boundary for the stack frame,
+and in fact is then equal to `prev_alloc_ptr`. This should mean that you now only need to reserve 2 slots
+in each stack frame for the parent frame's pointers.
```
ashelkovnykov, line 341:
```suggestion
and in fact is then equal to `prev_alloc_ptr`. This means that only two slots need to be reserved
in each stack frame for the parent frame's pointers.
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+
+However, before we could test whether the lightweight stack was empty by
+computing `stack_ptr == frame_ptr`. Now instead, we will need to save the
+initial location of `stack_ptr` in a reserved slot. This is because it is not a
+fixed offset from `frame_ptr` since the total number of slots in a given stack frame
+can vary. So we end up with the same number of reserved slots in the execution
+phase by putting the frame pointer at the boundary. If this is where it ended, this would arguably
+have been a better choice.
```
ashelkovnykov, line 349:
```suggestion
Now we can no longer test whether the lightweight stack is empty by computing `stack_ptr == frame_ptr`. Instead, we need to save the
initial location of `stack_ptr` in a reserved slot. This is because it is not a
fixed offset from `frame_ptr` since the total number of slots in a given stack frame
can vary. Therefore, we have the same number of reserved slots in the execution phase.
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+
+However, recall that in the cleanup phase, the reserved pointers for the parent
+frame are moved from being `frame_ptr`-relative slots to `alloc_ptr`-relative
+slots. So `prev_alloc_ptr` can no longer be calculated relative to
+`frame_ptr`, and instead needs to be saved in an `alloc_ptr`-relative slot.
+
+In the cleanup phase, since the number of slots is fixed at 3 (as opposed to
+varying as in the execution phase), the emptiness of the lightweight stack can
```
ashelkovnykov, line 357:
```suggestion
In the cleanup phase, since the number of slots is fixed at 3 (as opposed to
the varying number in the execution phase), the emptiness of the lightweight stack can
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+be determined by checking if its equal to one of the reserved slots (which one
+depends on how you order them). So we no longer need to save the initial value
```
ashelkovnykov, line 359:
```suggestion
be determined by checking if it's equal to one of the reserved slots.
Therefore we no longer need to save the initial value
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+of `stack_ptr` in the cleanup phase.
```
ashelkovnykov, line 360:
```suggestion
of `stack_ptr` in the cleanup phase. However, the frame pointer is now `alloc_ptr`-relative,
therefore we need to save the location of `prev_alloc_ptr` (since this was implicitly identical to `frame_ptr`).
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+
+Thus we would have two different sets of reserved pointers in the execution and
```
ashelkovnykov, line 362:
```suggestion
Thus, we have two different sets of reserved pointers between the execution and
```

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+cleanup phases:
+ - `prev_frame_ptr`, `prev_stack_ptr`, `init_stack_ptr` in the execution phase,
+   and
+ - `prev_frame_ptr`, `prev_stack_ptr`, `prev_alloc_ptr` in the cleanup phase.
+ 
+ Having which set of pointers are being saved change between phases is more
+ confusing for the programmer. Thus, we put the `frame_ptr` in the "middle", and
+ make the same `frame_ptr == stack_ptr` check to see whether the lightweight
+ stack is empty in both the execution and cleanup phases.
```
ashelkovnykov, line 371:
```
This adds complexity for the programmer with no space or performance benefit to the system.
```

ashelkovnykov (changes requested):
Forgot suggestions exist in my previous review. Used suggestions in places where I'm more certain of changes being required / preferable.

docs/split-stack.md
```diff
@@ -0,0 +1,371 @@
...
+#### Allocation pointer
+
+The allocation pointer is always initialized to be equal to the previous frame's
+stack pointer when a stack frame is pushed. This pointer is used to represent
+the extent of the "heap" for the current stack frame. The remaining free memory
+at a moment in time for the virtual machine is always precisely the region
+between the current stack frame's stack pointer and allocation pointer (at least
+during the execution phase). Thus the term "split stack".
+
+Every time an allocation is made or reclaimed, the allocation
```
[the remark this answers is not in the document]
eamsden: It's because objects live in the "heap"/allocation area. The lightweight stack and stack slots only hold fixed-size references or datums. The allocation area is for arbitrary-sized nouns (and bytecode...and HAMTs...and...)

drbeefsupreme:
I need to put a temporary pause on this, since my work on jet cold state has made it apparent that `Hamt::preserve` probably shouldn't use the lightweight stack. Once that's settled I'll come back to this.

eamsden:
This is pretty stale, we'll revisit docs for this later.

PR #68, opened 2023-08-10
cargo: murmur3 target specific commit

drbeefsupreme:
the name of murmur3_32_nocopy was changed to murmur3_32_of_slices but to avoid breaking things we're just going to target the last commit where its called murmur3_32_nocopy


eamsden:
@drbeefsupreme please do the opposite, please target the latest commit and update the name

drbeefsupreme:
Ok I thought that was what made more sense, I must have misunderstood you

eamsden (approved):
LGTM

PR #66, opened 2023-07-26
WIP: interpreter: use a lightweight stack for intra-formula traversal

drbeefsupreme:
Addresses #51

PR #75, opened 2023-08-30
interpreter: lightweight stacks and TCO

eamsden:
supersedes #66


eamsden:
closing in favor of #76 which uses a less-crazy entry type for the lightweight stack, removes the noun-encoding of nockwork, and is based on status as of 08f14b83183483ec2dffe164258e733b3bfe8a9d

Issue #51, opened 2023-05-05
treewalking interpreter: use a lightweight stack for intra-formula traversal


eamsden:
Presently the treewalking interpreter pushes a new 2stackz frame for every subformula. We should instead do something like what `jam` does and use a lightweight stack to traverse the formula tree, only pushing new frames for (non-tail) 2 and 9 evaluations.

eamsden:
https://github.com/urbit/ares/blob/7176f39cf725ff7f7678633abaf7380906ef19bc/rust/ares/src/serialization.rs#L176-L179

https://github.com/urbit/ares/blob/7176f39cf725ff7f7678633abaf7380906ef19bc/rust/ares/src/serialization.rs#L181-L184

https://github.com/urbit/ares/blob/7176f39cf725ff7f7678633abaf7380906ef19bc/rust/ares/src/serialization.rs#L201

eamsden:
Closing because done

PR #76, opened 2023-08-31
interpreter: use lightweight stack and do TCO

eamsden:
supersedes #66 and #75, addresses #51


drbeefsupreme (commented):
One nit: It's not clear to me why the names `vale`, `thee`, `kale`, and `diet` were chosen. The rest are self-explanatory. I see `vale` show up in codegen in this repo (though I don't know what it means there) so I assume that's related - are the other names from codegen as well?

I didn't try building a new version of Vere to try running the baby pill but I presume you got that worked out. I can run the `(shax 0x1)` jamfile fine though.

I tried to see whether it was actually doing TCO by feeding it a tail-optimized and non-tail-optimized factorial call. But there's so much going on I couldn't actually sort through it.

```
> =gat1 !.  |=  n=@ud
  ?:  =(n 1)
    1
  (mul n $(n (dec n)))
> =gat2 !.  |=  n=@ud
  =/  t=@ud  1
  |-
  ^-  @ud
  ?:  =(n 1)  t
  $(n (dec n), t (mul t n))
> =sub-bad [[gat1 5] 9 2 10 [6 0 3] 0 2]
> =sub-good [[gat2 5] 9 2 10 [6 0 3] 0 2]
> .fac-bad/atom sub-bad
> .fac-good/atom sub-good
```

I made these jamfiles on a ship with the sham jets put in for `hoon.hoon`. Surprisingly, the non-tail optimized version actually pushed a total of 126014 frames, while the tail optimized version pushed a total of 126196 frames. I don't think that's actually an indicator that anything is wrong, it just wasn't the result I expected.

I tried printing out intermediate results if they were an atom, and if I squint it looks right: https://pastebin.com/8YUjHe3T

Other than that LGTM. Using structs for each opcode and splitting static and dynamic hints makes things easier to follow than the previous iteration.

eamsden:
@drbeefsupreme 

> I made these jamfiles on a ship with the sham jets put in for hoon.hoon. Surprisingly, the non-tail optimized version actually pushed a total of 126014 frames, while the tail optimized version pushed a total of 126196 frames. I don't think that's actually an indicator that anything is wrong, it just wasn't the result I expected.

My guess is that those push counts are dominated by the calls to `mul` and `dec`. Can you repeat the test with an unjetted `dec`? It should be tail recursive, and a non-TCOd `dec` will push approximately <input> frames while a TCOd `dec` should push just a small constant number.

eamsden:
> One nit: It's not clear to me why the names vale, thee, kale, and diet were chosen. The rest are self-explanatory. I see vale show up in codegen in this repo (though I don't know what it means there) so I assume that's related - are the other names from codegen as well?

Sometimes there isn't a good name for a thing and so I invent one vaguely related to context:
- vale: anagram of "eval"
- kale: vale but for a (c/k)ore.
- thee: three but one letter less
- diet: anagram of edit

eamsden:
Also yes I have confirmed that the baby pill runs with this PR.

drbeefsupreme:
It was hard to write a non-TCO dec when increment is the only arithmetic operation, so here's what I tried with list repetition:

```hoon
> =gat-bad =>  ~  |=  a=@
    =|  b=@
    |-
    ?:  =(b a)  ~
    [5 $(b +(b))]
> =gat-good =>  ~  |=  a=@
    =|  b=@
    =/  c=*  ~
    |-
    ?:  =(a b)  c
    $(b +(b), c [5 c])
```

```
gat-bad
sample  |  frames
10          |  43
100        |  223
1000      |  2023

gat-good
sample  |  frames
10         |  40
100       |  128
1000     |  1029
```

The tail version is better but still not the relatively constant amount expected? I've only ever seen this stuff in theory, never actually counted frames, so maybe this is normal.

drbeefsupreme:
Also since we're trying out bigger jam files now, can we increase the memory for NockStack allocated in `main.rs`? A few times now I've been segfaulting (testing this PR and previously) and wondering what the hell I was doing wrong until I tried increasing the memory.

drbeefsupreme:
@eamsden pointed out that `unifying_equality()` was pushing a frame, but it doesn't allocate at all so it's not a big performance hit.

Decrementing the frame counter in `unifying_equality()` and we get 2 stack frames for the tail version and `n+2` stack frames for the non-tail version. Hooray!

eamsden:
After much tracing I discovered, and reviewed the discovery with @drbeefsupreme, that the accumulation of calls to push_frame is from unifying equality, since these tests do a Nock 5 on each loop. This is fine since unifying equality immediately pops its frame when complete, and does no copying, since it emits no allocated nouns (there is no preserve() call in unifying equality). I did discover that the outer formulas in newly-pushed activation frames were not properly recorded as being in tail-position, so with that fix we now get proper tail recursion.

@drbeefsupreme 
> i get 2 frames for the good version running 1000 and 1002 for the bad version

(`repeat_5_1000_tc.jam` and `repeat_5_1000.jam`)

drbeefsupreme (approved):

rust/ares/src/interpreter.rs
```diff
@@ -1,72 +1,226 @@
...
+    ComputeResult,
+    Done,
+}
+
+#[derive(Copy, Clone)]
+struct Nock11S {
+    todo: Todo11S,
+    tag: Atom,
+    body: Noun,
+}
```
ashelkovnykov, line 204: @eamsden Is there any reason you chose not to unify the two hints, like in Vere? It seems doable:
```
#[derive(Copy, Clone)]
#[repr(u8)]
enum Todo11 {
    ComputeHint,
    ComputeResult,
    Done,
}

#[derive(Copy, Clone)]
struct Nock11 {
    todo: Todo11,
    tag: Atom,
    hint: Option<Noun>,
    body: Noun,
}
```
Though I suppose you'd have to check that the following invariant holds:
```
hint.is_none() => todo != ComputeHint
```

ashelkovnykov (commented):
> Sometimes there isn't a good name for a thing and so I invent one vaguely related to context:
> 
> vale: anagram of "eval"
> kale: vale but for a (c/k)ore.
> thee: three but one letter less
> diet: anagram of edit

Numbers only run:
```
zero
ichi
deux
tres
four
five
seis
nana
huit
nine
diez
onze
```

rust/ares/src/interpreter.rs
eamsden: The code flows easier by separating the cases. This is also the representation I use in `$nomm`, fwiw

rust/ares/src/interpreter.rs
```diff
@@ -77,465 +231,482 @@ pub fn interpret(
...
+                        push_formula(stack, five.left, false);
+                    }
+                    Todo5::ComputeRightChild => {
+                        five.todo = Todo5::TestEquals;
+                        five.left = res;
+                        *stack.top() = NockWork::Work5(five);
+                        push_formula(stack, five.right, false);
+                    }
+                    Todo5::TestEquals => {
+                        let saved_value_ptr = &mut five.left as *mut Noun;
```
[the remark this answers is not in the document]
eamsden: Yeah it's unnecessary.
```suggestion
                        let saved_value_ptr = &mut five.left;
```

ashelkovnykov (approved):

PR #79, opened 2023-09-01
test_data: add (shax 0x1) jamfile

drbeefsupreme:
built on a ship that has the math.rs %sham jets


eamsden (approved):
Good test data to have

PR #80, opened 2023-09-01
main: make assert_no_alloc actually assert

eamsden:
Can be tested with the patch:

```patch
diff --git a/rust/ares/src/jets/math.rs b/rust/ares/src/jets/math.rs
index 9993cb0..2882291 100644
--- a/rust/ares/src/jets/math.rs
+++ b/rust/ares/src/jets/math.rs
@@ -67,6 +67,16 @@ pub fn jet_add(stack: &mut NockStack, subject: Noun) -> Result<Noun, JetErr> {
     let a = raw_slot(arg, 2).as_atom()?;
     let b = raw_slot(arg, 3).as_atom()?;
 
+    eprintln!("AAAAAHHHH I'M ALLOOOOOOCATING");
+
+    let mut v : Vec<u64> = Vec::with_capacity(0);
+
+    v.push(1);
+    v.push(2);
+    v.push(3);
+
+    eprintln!("ALLOOOOOOCATED. Element 3: {}", v[2]);
+
     if let (Ok(a), Ok(b)) = (a.as_direct(), b.as_direct()) {
         Ok(Atom::new(stack, a.data() + b.data()).as_noun())
     } else {
```

### DIFF
--- a/rust/ares/src/interpreter.rs
+++ b/rust/ares/src/interpreter.rs
@@ -83,6 +83,17 @@ pub fn interpret(
         *(stack.local_noun_pointer(0)) = work_to_noun(Done);
     }
     push_formula(stack, formula);
+    // DO NOT REMOVE THIS ASSERTION
+    //
+    // If you need to allocate for debugging, wrap the debugging code in
+    //
+    // ```
+    // permit_alloc(|| {
+    //   your.code.goes.here()
+    // })
+    // ```
+    //
+    // (See https://docs.rs/assert_no_alloc/latest/assert_no_alloc/#advanced-use)
     assert_no_alloc(|| unsafe {
         loop {
             match noun_to_work(*(stack.local_noun_pointer(0))) {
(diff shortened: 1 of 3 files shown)


philipcmonk:
For context, this was in lib.rs before: https://github.com/urbit/ares/blob/c26fd83e2787a895c8e0236848bd0f25b6bcc02a/rust/ares/src/lib.rs#L39-L41.  I have no opinion on where it actually lives.

IIRC I commented that out so that I could printf in jets.

eamsden:
FFR: https://docs.rs/assert_no_alloc/latest/assert_no_alloc/#advanced-use

use `permit_alloc()` to allow allocation in an assert_no_alloc. I will make a comment about this at the assertion and allocator sites.

drbeefsupreme:
This works - trying to allocate a vector in Ares called as a jet from Vere:

```
~dev:dojo> ;;([bpoly:f bpoly:f] (jinx-gate:f bpdvr:f bp3 bp2))memory allocation of 7992 bytes failed
pier: serf unexpectedly shut down
```

drbeefsupreme (approved):

PR #84, opened 2023-09-13
hamt: preserve() traversal_stack uses rust stack

drbeefsupreme:
This was using the lightweight stack, but this presented the following issue: `NockStack::copy()` can be called during the `Hamt::preserve()` call, and the lightweight stack may not be empty. `NockStack::copy()` presumes that the lightweight stack is empty when it is called, and thus tried to pop elements that `preserve()` had pushed onto it.

`NockStack::copy()` should probably assert that the lightweight stack is empty when its called.

I'm making the assert `stack_is_empty()` a separate PR #85 since its a separate concern, but I have tested that it works fine with this PR and several jamfiles where `copy()` is called many times.

### DIFF
--- a/rust/ares/src/hamt.rs
+++ b/rust/ares/src/hamt.rs
@@ -451,9 +451,8 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
                 if traversal_depth == 0 {
                     break;
                 }
-                let Some((stem, mut position)) = traversal_stack[traversal_depth - 1] else {
-                    panic!("Attempted to access uninitialized array element");
-                };
+                let (stem, mut position) = traversal_stack[traversal_depth - 1]
+                    .expect("Attempted to access uninitialized array element");
                 // can we loop over the size and count leading 0s remaining in the bitmap?
                 'preserve_stem: loop {
                     if position >= 32 {
--- a/rust/ares/src/hamt.rs
+++ b/rust/ares/src/hamt.rs
@@ -437,17 +437,26 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
             let dest_buffer = stack.struct_alloc_in_previous_frame(self.0.size());
             copy_nonoverlapping(self.0.buffer, dest_buffer, self.0.size());
             self.0.buffer = dest_buffer;
-            *(stack.push::<(Stem<T>, u32)>()) = (self.0, 0);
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
Hold off on this one moment... I thought I tested this with my jet cold state branch, but I must have made a mistake somewhere since now its not working.

drbeefsupreme:
Nevermind, all good. I just forgot that I had some of the changes in the jet cold state branch I needed stashed, not committed, and its working with the stash restored.

ashelkovnykov:
Haven't looked at the code in this PR yet, but I just found that `%memo` hints were causing Ares crashes on `preserve` calls to the cache in my stack trace test pill. As another data point, I'm going to try applying this PR to my branch to see if it fixes the crashes.

rust/ares/src/hamt.rs
```diff
@@ -437,17 +437,26 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+            traversal_stack[0] = Some((self.0, 0));
             let mut traversal_depth = 1;
             'preserve: loop {
                 if traversal_depth == 0 {
                     break;
                 }
-                let (stem, mut position) = *(stack.top::<(Stem<T>, u32)>());
+                let Some((stem, mut position)) = traversal_stack[traversal_depth - 1] else {
+                    panic!("Attempted to access uninitialized array element");
+                };
```
eamsden, line 456:
```suggestion
                let (stem, mut position) = traversal_stack[traversal_depth - 1].expect("Attempted to access uninitialized array element");
```

`.expect()` is the canonical way to do this for both `Result` and `Option`

eamsden (changes requested):
Just one change for idiomatic/succinct Rust.

ashelkovnykov:
> Haven't looked at the code in this PR yet, but I just found that `%memo` hints were causing Ares crashes on `preserve` calls to the cache in my stack trace test pill. As another data point, I'm going to try applying this PR to my branch to see if it fixes the crashes.

Confirmed that this PR fixes `%memo` hints in my stack traces pill

eamsden (approved):

ashelkovnykov (commented):
Late, but LGTM.

The code all looks logically correct, though I don't understand the Ares `Hamt` code fully (for lack of trying).

PR #85, opened 2023-09-13
mem: assert stack_is_empty() when it should be

drbeefsupreme:
if these functions are callled with a non-empty lightweight stack, they will try to pop things that they shouldn't.

all other stack_is_empty() callsites happen in a fresh stack frame, so we already know the lightweight stack will be empty.

This would have made detecting the problem that #84 resolves take a few minutes instead of a few hours.


eamsden (approved):
Good assertions are good.

PR #89, opened 2023-09-20
codegen: SACK analysis and new linearizer

eamsden:
Supersedes #54 since the history of that was far too messy

**Original**

This PR upgrades SKA to SACK, which is capable of discerning which knowledge is actually necessary to fix called nock formulas for direct calls, and use this information as the battery label for generated code.

This avoids the TMI problem when using SKA to build cell structures in accumulators, and the battery discovery problem which required several ad-hoc heuristics to discover stable batteries for dynamic cores such as Arvo, vanes, gall agents, threads, generators, and dojo expressions. It also allows us to directly derive "cold state"[1] for jetting from fast hints, and to statically link jets using no more information than is already used for the current vere jetting system.


eamsden:
Closing in favor of #135

PR #90, opened 2023-09-22
build: add `urcrypt`

matthew-levan:
Adds `urcrypt-sys`, a crate that provides Rust bindings to `urcrypt`, which is now required as a system dependency for building Ares. Also adds dependency installation commands in the CI script.


eamsden (changes requested):
Marker review to make sure we audit the licenses of everything going into this codebase. @matthew-levan please reply to this review comment with a checklist of vendored libraries and a list of their licenses.

rust/shell.nix
```diff
@@ -14,5 +14,6 @@ pkgs.mkShell {
     ])
     cargo-watch
     gdb
+    secp256k1
```
eamsden, line 17: `urcrypt` should be built in the Nix environment. We should not require global installs of any package for developers.

eamsden (changes requested):

eamsden:
I'll leave this open for now but it's likely to close in favor of #123 shortly.

eamsden:
Closing in favor of #123 (which adds urcrypt to the nix shell so that it is easy for local devs to build ares with urcrypt) and #124 (which fixes a bug in `jet_puck()`)

PR #95, opened 2023-09-28
jets: add `trip`

matthew-levan:


eamsden (changes requested):
I don't think this is necessary, it's just a special case of the `+rip` jet.

matthew-levan:
Sounds good, wonder why it's jetted in Vere though.

eamsden:
Even if it's unjetted, as long as there are jets for the actual bit-mashing operations it should be fine.

joemfb:
Lots of things are jetted in vere just to provide the api to other code -- once we've written the function, might as well wire up the jet too. (`+need` is probably the best example of this.)

matthew-levan:
Ah, that makes sense. Thanks. If we need `crate::jets::util::trip` for
another jet down the line, we can restore this branch I guess.

PR #99, opened 2023-10-03
Fix loading from snapshot

philipcmonk:
`write_bytes` takes a number of word-sized entries, not a number of bytes.


ashelkovnykov (approved):

eamsden:
Yes this is an annoyingly badly named entry in rust's stdlib

PR #101, opened 2023-10-03
interpreter: permit_alloc for slogs

drbeefsupreme:
Simple as.

### DIFF
--- a/rust/ares/src/interpreter.rs
+++ b/rust/ares/src/interpreter.rs
@@ -5,7 +5,7 @@ use crate::mem::NockStack;
 use crate::newt::Newt;
 use crate::noun::{Atom, Cell, DirectAtom, IndirectAtom, Noun, D, T};
 use ares_macros::tas;
-use assert_no_alloc::assert_no_alloc;
+use assert_no_alloc::{assert_no_alloc, permit_alloc};
 use bitvec::prelude::{BitSlice, Lsb0};
 use either::Either::*;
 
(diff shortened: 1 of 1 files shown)


eamsden (changes requested):
We need to figure out how to do this without the rust allocator.

Pretty sure `println!()` allocates to do the formatting. Maybe we need basic string formatting on the stack allocator.

drbeefsupreme:
So I understand why we don't want to use the Rust allocator for anything to do with the `NockStack`. But why do we need to worry about it for side effects like printing?

eamsden:
We don't want to do it from anywhere within `interpret()` because we might want to terminate an thread running interpret() without unwinding the stack

ashelkovnykov:
> @ashelkovnykov would like your opinion on this

@philipcmonk and I talked about this before. The ideal case would be versions of whatever trait it is that handles `string` formatting that allocate on the `NockStack` (well, *actually* the ideal case would be either Rust doing formatting on the stack or the Rust compiler realizing that this allocation is immediately flushed and deallocated).

On the one hand, I think this is fine because it's only going to be used for debugging and the allocation is deallocated immediately (unless you crash during the `println` in which case you have bigger issues). On the other hand, I don't think the case of not having a `Newt` even makes sense any more, now that we can reliably hook up `ares` to `king`.

Before `ares` is production ready, we're going to both remove this codepath *and* have `NockStack` allocators for formatting, therefore I don't feel strongly about this change. Your call @eamsden .

drbeefsupreme:
I'm not sure its right that the buffer for `println` is immediately deallocated. You can put a `println` in `interpret()` before `assert_no_alloc`, and then `println`s within `assert_no_alloc` work perfectly fine, suggesting they're using the same buffer that was allocated from the first `println`.

eamsden:
The ability to run `interpret()` without a newt is going to be removed post-Assembly, so this PR would be quickly obsoleted.

PR #98, opened 2023-10-03
Fix error in senior_pointer_first

ashelkovnykov:
courtesy of @joemfb. The stuff of nightmares.

PR #103, opened 2023-10-04
Further fixes to `senior_pointer_first()` and stack initialization

ashelkovnykov:
See [here](https://github.com/urbit/ares/pull/98#issuecomment-1745907804).

### DIFF
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -70,9 +70,9 @@ impl NockStack {
         let stack_pointer = frame_pointer;
         let alloc_pointer = unsafe { start.add(size) } as *mut u64;
         unsafe {
-            *frame_pointer = ptr::null::<u64>() as u64; // "frame pointer" from "previous" frame
+            *frame_pointer.sub(FRAME) = ptr::null::<u64>() as u64; // "frame pointer" from "previous" frame
             *frame_pointer.sub(STACK) = ptr::null::<u64>() as u64; // "stack pointer" from "previous" frame
-            *frame_pointer.sub(ALLOC) = start as u64; // "alloc pointer" from "previous" frame
+            *frame_pointer.sub(ALLOC) = ptr::null::<u64>() as u64; // "alloc pointer" from "previous" frame
         };
         NockStack {
             start,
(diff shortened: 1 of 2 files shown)


ashelkovnykov:
Need to run now, but realizing that I should add the "pick lower address" logic to the `break` case.

rust/ares/src/mem.rs
```diff
@@ -805,6 +805,11 @@ unsafe fn senior_pointer_first(
     };
 
     loop {
+        if low_pointer.is_null() || high_pointer.is_null() {
+            // we found the top of the stack
+            break (a, b); // both are in the bottom frame, pick arbitrarily
```
ashelkovnykov, line 810: Noting for myself that I should copy the logic from lines `819` to `823`.

rust/ares/src/mem.rs
```diff
@@ -70,9 +70,9 @@ impl NockStack {
         let stack_pointer = frame_pointer;
         let alloc_pointer = unsafe { start.add(size) } as *mut u64;
         unsafe {
-            *frame_pointer = ptr::null::<u64>() as u64; // "frame pointer" from "previous" frame
+            *frame_pointer.sub(FRAME) = ptr::null::<u64>() as u64; // "frame pointer" from "previous" frame
```
eamsden, line 73: these are all wrong no? They should all be `.sub(CONSTANT + 1)`?

rust/ares/src/mem.rs
eamsden, line 810: As noted in @ashelkovnykov's comment this should return the lower of the two pointers.

eamsden (changes requested):

rust/ares/src/mem.rs
eamsden: Also, the comment above mis-states the test for a west frame, so we should correct it.

eamsden: I think if the above is corrected then the extra null pointer check is unneeded.

eamsden (changes requested):

eamsden (changes requested):

rust/ares/src/mem.rs
eamsden: (nevermind this is wrong, you do still need it)

ashelkovnykov: > Also, the comment above mis-states the test for a west frame, so we should correct it.

What's incorrect about this? `frame < alloc` is a valid and correct test for a West frame

eamsden (approved):

PR #104, opened 2023-10-04
stack: write null pointers to proper slots

eamsden:
@ashelkovnykov I think this fixes it?


rust/ares/src/mem.rs
```diff
@@ -805,11 +805,6 @@ unsafe fn senior_pointer_first(
     };
 
     loop {
-        if low_pointer.is_null() || high_pointer.is_null() {
-            // we found the top of the stack
-            break (a, b); // both are in the bottom frame, pick arbitrarily
-        }
```
eamsden, line 811: Actually it's even different than this: if the previous stack pointer is null we need to use `stack.start.add(stack.size)` and check if in frame, so that we properly distinguish seniority between the stack and e.g. the PMA

PR #105, opened 2023-10-04
Add SIGINT handling to Ares

ashelkovnykov:
- Add simple `SIGINT` handling:
  - on `SIGINT` from `king`, set flag value
  - check flag before Nock `2` frame push, Nock `9` frame push, and mean stack push
  - bail out if flag set with `NonDeterministic` error
  - a second `SIGINT` while  the flag is set kills the process immediately
- Forward `NonDeterministic` errors to the most senior frame
- modified `toddler.hoon` for easier interrupt testing
- ordered Cargo imports alphabetically
- disabled checks for cyclic `Noun`s at Cargo level

### DIFF
--- a/rust/ares/Cargo.toml
+++ b/rust/ares/Cargo.toml
@@ -42,4 +42,3 @@ opt-level = 3
 opt-level = 3
 
 [features]
-check_acyclic=[]
--- a/rust/ares/Cargo.toml
+++ b/rust/ares/Cargo.toml
@@ -11,18 +11,20 @@ edition = "2018"
 
 [dependencies]
 ares_macros = { path = "../ares_macros" }
-bitvec = "1.0.0"
-either = "1.9.0"
-libc = "0.2.126"
-murmur3 = { git = "https://github.com/tloncorp/murmur3", rev = "7878a0f" }
-memmap = "0.7.0"
-intmap = "1.1.0"
-num-traits = "0.2"
-num-derive = "0.3"
-criterion = "0.4"
-static_assertions = "1.1.0"
-ibig = { path = "../ibig-rs" }
 assert_no_alloc = "1.1.2"
+bitvec = "1.0.0"
+criterion = "0.4"
+either = "1.9.0"
+ibig = { path = "../ibig-rs" }
+intmap = "1.1.0"
+lazy_static = "1.4.0"
+libc = "0.2.126"
+memmap = "0.7.0"
+murmur3 = { git = "https://github.com/tloncorp/murmur3", rev = "7878a0f" }
+num-derive = "0.3"
+num-traits = "0.2"
+signal-hook = "0.3"
+static_assertions = "1.1.0"
 
 [build-dependencies]
 cc = "1.0.79"
--- a/rust/ares/Cargo.toml
+++ b/rust/ares/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2018"
 # [patch.crates-io]
 # ibig = { path = "../ibig-rs" }
 
+# Please keep these alphabetized
 [dependencies]
 ares_macros = { path = "../ares_macros" }
 assert_no_alloc = "1.1.2"
--- a/rust/ares/Cargo.toml
+++ b/rust/ares/Cargo.toml
@@ -12,6 +12,8 @@ edition = "2018"
 [dependencies]
 ares_macros = { path = "../ares_macros" }
 assert_no_alloc = "1.1.2"
+# use this when debugging requires allocation (e.g. eprintln)
+# assert_no_alloc = {version="1.1.2", features=["warn_debug"]}
 bitvec = "1.0.0"
 criterion = "0.4"
 either = "1.9.0"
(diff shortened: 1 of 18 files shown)


rust/ares/Cargo.toml
```diff
@@ -11,20 +11,22 @@ edition = "2018"
...
 libc = "0.2.126"
-murmur3 = { git = "https://github.com/tloncorp/murmur3", rev = "7878a0f" }
 memmap = "0.7.0"
-intmap = "1.1.0"
-num-traits = "0.2"
+murmur3 = { git = "https://github.com/tloncorp/murmur3", rev = "7878a0f" }
 num-derive = "0.3"
-criterion = "0.4"
+num-traits = "0.2"
+signal-hook = "0.3"
```
eamsden, line 28: was this reordering done by some automated tool? Or just manually alphabetized?

rust/ares/src/hamt.rs
```diff
@@ -480,7 +480,7 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
                                 };
                                 *(stem.buffer.add(idx) as *mut Entry<T>) = Entry { stem: new_stem };
                                 assert!(traversal_depth <= 5); // will increment
-                                traversal_stack[traversal_depth - 1].unwrap().1 = position + 1;
+                                traversal_stack[traversal_depth - 1] = Some((stem, position + 1));
```
eamsden, line 483: Marker for synchronous review: question about this change.

rust/ares/src/jets/nock.rs
```diff
@@ -22,12 +23,15 @@ pub fn jet_mink(
     let v_formula = slot(arg, 5)?;
     let _scry = slot(arg, 3)?;
 
-    //  XX: NonDeterministic errors need to bail here, too
+    //  XX: no partial traces; all of our traces go down to the "home road"
```
eamsden, line 26: So we are switching from "Deterministic and non-deterministic traces both trace only to the nearest virtualization layer" to "Deterministic and non-deterministic traces both trace all the way to the outermost frame"?

rust/ares/src/jets/nock.rs
```diff
@@ -252,6 +256,21 @@ pub mod util {
...
+    fn init() {
+        // This needs to be done because TERMINATOR is lazy allocated, and if you don't
+        // do it before you call the unit tests it'll get allocated on the Rust heap
+        // inside an assert_no_alloc block.
+        //
+        // Also Rust has no primitive for pre-test setup / post-test teardown, so we
+        // do it in a test that we rely on being called before any other in this file,
+        // since we're already using single-threaded test mode to avoid race conditions
+        // (because Rust doesn't support test order dependencies either).
+        let _ = Arc::clone(&TERMINATOR);
```
eamsden, line 272: Ew. That's all I can say.

rust/ares/src/mem.rs
```diff
@@ -528,6 +528,16 @@ impl NockStack {
...
+        if self.frame_pointer.is_null()
+            || self.stack_pointer.is_null()
+            || self.alloc_pointer.is_null()
+        {
+            panic!(
+                "serf: frame_pop: null NockStack pointer f={:p} s={:p} a={:p}",
+                self.frame_pointer, self.stack_pointer, self.alloc_pointer
+            );
+        }
+
```
eamsden, line 540: panic!s can be wrapped in `permit_alloc()` since we know they will unwind the rust stack.

rust/ares/src/serf.rs
```diff
@@ -23,12 +25,20 @@ const POKE_AXIS: u64 = 23;
...
+    pub static ref TERMINATOR: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
+}
+
 /**
  * This is suitable for talking to the king process.  To test, change the arg_c[0] line in
  * u3_lord_init in vere to point at this binary and start vere like normal.
  */
 pub fn serf() -> io::Result<()> {
-    sleep(time::Duration::from_secs(0));
+    // Register SIGTERM signal hook to set flag first time, shutdown second time
```
eamsden, line 38:
```suggestion
    // Register SIGINT signal hook to set flag first time, shutdown second time
```

eamsden (commented):
Will review synchronously

rust/ares/src/jets/nock.rs
ashelkovnykov: No, this is just documenting that all traces do not stop at virtualization layers. This was the case before this PR as well, it's just especially relevant now that interrupts are available.

rust/ares/src/hamt.rs
ashelkovnykov: This is fixing a logical error that was not causing a bug due to safety logic up above (we discovered it together during the very first pairing session for the memory seniority bug):

The `lvalue` here is incorrect and not actually resulting in the `Stem` in the path at depth `traversal_depth - 1` to be updated. The compiler is moving the value and inferring mutability to the moved value, which is then updated, leaving the original untouched. However, when we come back around with the same `position` value as before, the `buffer` of the `Stem` has already been copied into the parent frame, resulting in the `is_in_frame()` check to return false, and therefore immediately incrementing the `position` value and continuing the loop.

rust/ares/Cargo.toml
ashelkovnykov: This was manual (because of the 'tism).

eamsden (approved):

PR #106, opened 2023-10-05
jets: add `dor`, `mor`, `gor`

matthew-levan:


rust/ares/src/jets/order.rs
```diff
@@ -0,0 +1,151 @@
...
+                if b.is_atom() {
+                    Ok(NO)
+                } else {
+                    let a = slot(a, 2).unwrap();
+                    let b = slot(b, 2).unwrap();
+                    if unsafe { a.raw_equals(b) } {
+                        Ok(dor(stack, slot(a, 3)?, slot(b, 3)?)?)
+                    } else {
+                        Ok(dor(stack, slot(a, 2)?, slot(b, 2)?)?)
+                    }
```
joemfb, line 101: This isn't right. You're replacing `a` and `b` with their respective heads, then again taking head or tail when branching on their equality, which means the recursive calls will operate on the head of the head or the tail of the head.

matthew-levan: Thank you.

rust/ares/src/jets/math.rs
```diff
@@ -303,6 +303,29 @@ pub fn jet_sub(
     Ok(sub(stack, a, b)?.as_noun())
 }
 
+pub mod util {
+    use crate::noun::{Atom, Noun, YES, NO};
+    use crate::mem::NockStack;
+
+    pub fn lth(stack: &mut NockStack, a: Atom, b: Atom) -> Noun {
```
ashelkovnykov, line 310: 1. This isn't removing the logic from `jet_lth`, it's just cloning it into a second place
2. A util `lth` function should not take a `NockStack` since it never allocates
3. I've been inconsistent thus far about where shared functions go: in `jets::util` or in `jets::[module]::util`. We should probably pick which one is canonical and stick with it from now on.

rust/ares/src/jets.rs
```diff
@@ -3,6 +3,7 @@ pub mod form;
 pub mod hash;
 pub mod math;
 pub mod nock;
+pub mod order;
```
ashelkovnykov, line 6: Nit: not 4 characters. Also, I would have expected these to go into `hash.rs`.

rust/ares/src/jets/order.rs
```diff
@@ -0,0 +1,153 @@
+use crate::jets::util::slot;
```
ashelkovnykov, line 1: Style error imported from the file you copied: import above file descriptor comment

rust/ares/src/jets/order.rs
```diff
@@ -0,0 +1,153 @@
...
+/** Text processing jets
+ */
```
ashelkovnykov, line 3: Forgot to update the file description

rust/ares/src/jets/order.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::jets;
+use crate::jets::order::util::dor;
```
ashelkovnykov, line 5: This is a weird import, since this function is defined within this very file. I'm not actually sure what the canonical Rust style is in this case. I think it might be just prefixing the three call sites with `util::`.

rust/ares/src/jets/order.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+pub mod util {
+    use crate::jets::util::slot;
+    use crate::jets::math::util::lth;
+    use crate::jets::Result;
+    use crate::mem::NockStack;
+    use crate::noun::{Noun, YES, NO};
+
+    pub fn dor(stack: &mut NockStack, a: Noun, b: Noun) -> Result {
```
ashelkovnykov, line 81: `dor` doesn't need to be passed a `NockStack` because neither it nor any function it calls ever allocate

ashelkovnykov (changes requested):

rust/ares/src/jets/math.rs
matthew-levan: 3. `jets::math::util::lth` requires a `NockStack` because it uses the `Atom::as_ubig` function. Is there another function I could use instead, which doesn't require a stack?

rust/ares/src/jets.rs
matthew-levan: How about `sort`?

rust/ares/src/jets/math.rs
ashelkovnykov: Yeah, you're right - I just tried replacing those calls with `as_bitslice()` and comparing those, but the comparator for `BitSlice`s is broken.

rust/ares/src/jets/math.rs
```diff
@@ -303,6 +289,30 @@ pub fn jet_sub(
     Ok(sub(stack, a, b)?.as_noun())
 }
 
+pub mod util {
+    use crate::jets::Result;
+    use crate::noun::{Atom, YES, NO};
+    use crate::mem::NockStack;
+
+    pub fn lth(stack: &mut NockStack, a: Atom, b: Atom) -> Result {
```
ashelkovnykov, line 297: The `util` version of `lth` should probably return a `bool` and let the `jet_lth` wrapper handle returning a `jets::Result`

rust/ares/src/jets/sort.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+pub mod util {
+    use crate::jets::util::slot;
+    use crate::jets::math::util::lth;
+    use crate::jets::Result;
+    use crate::mem::NockStack;
+    use crate::noun::{Noun, YES, NO};
+
+    pub fn dor(stack: &mut NockStack, a: Noun, b: Noun) -> Result {
```
ashelkovnykov, line 80: Same as above - the `util` version should probably return `bool`

rust/ares/src/jets/sort.rs
```diff
@@ -0,0 +1,152 @@
...
+                    let b_head = slot(b, 2)?;
+                    let a_tail = slot(a, 3)?;
+                    let b_tail = slot(b, 3)?;
+                    if unsafe { a_head.raw_equals(b_head) } {
+                        Ok(dor(stack, a_tail, b_tail)?)
+                    } else {
+                        Ok(dor(stack, a_head, b_head)?)
+                    }
+                }
+            }
```
ashelkovnykov, line 104: This is unidiomatic in Rust. The preferred syntax is:
```
match (a.either(), b.either()) {
    (Left(atom_a) Left(atom_b)) => lth(stack, atom_a, atom_b),
    (Left(_), Right(_)) => true,
    (Right(_), Left(_)) => false,
    (Right(cell_a), Right(cell_b)) => {
        let a_head = slot(cell_a, 2)?;
        let b_head = slot(cell_b, 2)?;
        let a_tail = slot(cell_a, 3)?;
        let b_tail = slot(cell_b, 3)?;
        if unsafe { a_head.raw_equals(b_head) } {
            dor(stack, a_tail, b_tail)
        } else {
            dor(stack, a_head, b_head)
        }
    }
}
```

rust/ares/src/jets/math.rs
matthew-levan: That sounds good, except how should we decide on whether to return a `bool` or loobean? It doesn't seem like we've standardized on this yet, across the rest of the codebase.

ashelkovnykov (approved):

rust/ares/src/jets/math.rs
ashelkovnykov: Yeah, there's no standard for utility functions so far, other than "what's easiest to work with for most of the functions calling the `util` function? So by that metric, I think returning a `loobean` `Atom` for `lth` and `dor` is fine.

joemfb: I think it's important to set a standard and stick to it. Booleans are obviously manageable -- loobeans actually are too. Switching between them unpredictably in C is not at all. But with rust's stronger type system it might be ...

/cc @eamsden

PR #110, opened 2023-10-10
Add debug assertions from NockStack investigation

ashelkovnykov:
Adds assertion macros controlled by Rust feature flags to vital functions in `interpreter.rs`, `mem.rs`, and `mug.rs`. These assertions were useful for debugging during the `NockStack` incident.

@eamsden Perhaps they should all be enabled for CI?


eamsden:
> Perhaps they should all be enabled for CI?
@ashelkovnykov yes please

eamsden (approved):

PR #111, opened 2023-10-11
Handle scries in Ares with virtual Nock 12

ashelkovnykov:


rust/ares/src/interpreter.rs
```diff
@@ -637,6 +658,67 @@ pub fn interpret(context: &mut Context, mut subject: Noun, formula: Noun) -> Res
...
+                                ],
+                            );
+                            let scry_form = T(context.stack, &[D(9), D(2), D(1), scry_core]);
+
+                            scry.todo = Todo12::Done;
+                            scry.path = res;
+                            scry.hand = context.scry_stack;
+                            context.scry_stack = cell.tail();
+                            *context.stack.top() = NockWork::Work12(scry);
+                            push_formula(context.stack, scry_form, false)?;
```
eamsden, line 693: Will this crash in the correct way? The scry gate does not run in the current virtualized context but in the previous one, and therefore we want the outer rather than inner interpret call to return NockErr if it crashes. @philipcmonk @ashelkovnykov

rust/ares/src/mem.rs
```diff
@@ -979,6 +979,77 @@ pub unsafe fn unifying_equality(stack: &mut NockStack, a: *mut Noun, b: *mut Nou
     (*a).raw_equals(*b)
 }
 
+pub fn non_unifying_equality(a: Noun, b: Noun) -> bool {
```
philipcmonk, line 982: This doesn't handle the case of densely-connected DAGs, where the same expensive comparison might happen many times.  See `_cr_sing_cape` in vere for one approach to this, which is to keep a set of comparisons we've already made.  Unifying equality doesn't need this, because it unifies as it goes.

However, since the goal of this function is primarily to operate on Nouns directly instead of mutable references, it probably should just create a mutable reference on its own stack by copying the noun and then call regular unifying equality.

rust/ares/src/interpreter.rs
```diff
@@ -925,48 +1026,71 @@ fn push_formula(stack: &mut NockStack, formula: Noun, tail: bool) -> Result<(),
...
+                T(stack, &[h, t])
+            }
+        };
+
+        while (stack).get_frame_pointer() != virtual_frame {
+            (stack).preserve(&mut preserve);
+            (stack).frame_pop();
+        }
+
+        match error {
```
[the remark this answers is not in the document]
ashelkovnykov: It ended up being uglier to deal with this than having a second `match` statement, so I've left it alone for now.

eamsden (approved):
I really appreciate the generous comment about how scry crashes are unwound. Thanks @ashelkovnykov

PR #113, opened 2023-10-18
Revert `%memo` cache to "per road"

ashelkovnykov:
Requires #111 to be merged first

### DIFF
--- a/rust/ares/src/interpreter.rs
+++ b/rust/ares/src/interpreter.rs
@@ -258,8 +258,6 @@ pub struct Context<'a> {
     pub stack: &'a mut NockStack,
     // For printing slogs; if None, print to stdout; Option slated to be removed
     pub newt: Option<&'a mut Newt>,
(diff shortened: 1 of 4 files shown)


ashelkovnykov:
@eamsden Retargeted

rust/ares/src/interpreter.rs
```diff
@@ -258,8 +258,6 @@ pub struct Context<'a> {
     pub stack: &'a mut NockStack,
     // For printing slogs; if None, print to stdout; Option slated to be removed
     pub newt: Option<&'a mut Newt>,
-    // Per-event cache; option to share cache with virtualized events
-    pub cache: &'a mut Hamt<Noun>,
```
eamsden, line 262: We need to keep this in the context and just reset it, as the `ut` jets need it.

eamsden (changes requested):

PR #115, opened 2023-10-20
++xeb, ++flop

sigilante:


sigilante:
`++swp` is commented out here since it seems to not matter much.  (The Vere is inefficient too so the other jets it uses should be enough.)

eamsden (approved):

PR #116, opened 2023-10-20
Post jam/cue jets.

sigilante:
Should probably include some explicit round-trip tests rather than just hard-coded ones too.


rust/ares/src/jets/serial.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::noun::{Atom, DirectAtom, IndirectAtom, Noun, D, DIRECT_MAX, NO, T, YES};
+use crate::mem::NockStack;
+use crate::serialization::{cue, jam};
+
+crate::gdb!();
+
+pub fn jet_cue(context: &mut Context, subject: Noun) -> Result {
+    let stack = &mut context.stack;
+    let sam = slot(subject, 6)?;
+    let atom = slot(sam, 1)?.as_atom()?;
```
eamsden, line 14: this is unnecessary: slot(x, 1) is the identity function on nouns.

```suggestion
    let atom = sam.as_atom()?;
```

rust/ares/src/jets/serial.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    Ok(cue(stack, atom))
+}
+
+pub fn jet_jam(context: &mut Context, subject: Noun) -> Result {
+    let stack = &mut context.stack;
+    let sam = slot(subject, 6)?;
+    let noun = slot(sam, 1)?;
+
+    Ok(jam(stack, noun).as_noun())
```
eamsden, line 24: as above

```suggestion
    Ok(jam(stack, sam).as_noun())
```

eamsden (changes requested):

eamsden (approved):
LGTM

PR #117, opened 2023-10-25
Jets for `min`, `max`, and `peg`

ashelkovnykov:


rust/ares/Cargo.toml
```diff
@@ -48,4 +48,4 @@ check_all = [ "check_acyclic", "check_forwarding", "check_junior" ]
 check_acyclic = []
 check_forwarding = []
 check_junior = []
-sham_hints = []
+sham_hints = []
```
ashelkovnykov, line 51: Merge error

rust/ares/src/jets/bits.rs
```diff
@@ -525,34 +537,33 @@ mod tests {
...
-        let sam = T(&mut c.stack, &[bit, a24]);
-        assert_jet(c, jet_lsh, sam, D(0x10eca860));
-        let bit = T(&mut c.stack, &[D(4), D(6)]);
-        let sam = T(&mut c.stack, &[bit, a96]);
-        let res = A(
-            &mut c.stack,
-            &ubig!(_0xfaceb00c15deadbeef123456000000000000000000000000),
-        );
-        assert_jet(c, jet_lsh, sam, res);
+        // let (a0, a24, _a63, a96, a128) = atoms(&mut c.stack);
```
ashelkovnykov, line 540: Forgot to uncomment tests

rust/ares/src/jets/tree.rs
```diff
@@ -40,13 +42,71 @@ pub fn jet_mas(context: &mut Context, subject: Noun) -> Result {
     }
 }
 
+pub fn jet_peg(context: &mut Context, subject: Noun) -> Result {
```
eamsden, line 45: in general this would do less copying, and arguably be clearer, if it were written in terms of bitslices. The idea is to allocate an indirect atom that can hold `(sub 1 (add (met 0 a) (met 0 b)))` bits, copy all but the MSB of `b` in, then copy all of `a` in. We omit the MSB of `b` since the MSB of an axis is the size marker, and all bits following are the actual path.

eamsden (changes requested):

eamsden:
I rewrote `jet_peg` and `jet_mas` to use bitslices: 4177ae6

rust/ares/src/jets/tree.rs
eamsden: See 4177ae6 / `eamsden/peg-mas-bitslice` branch

ashelkovnykov: You're right. The previous version was written using the existing jets as reference material, which at the time did not have @joemfb 's rewrite of the `+cap`, `+mas`, and `+peg` jets in [397](https://github.com/urbit/vere/pull/397) merged. I was also ignorant at the time of the relationship between the aforementioned jets and the bits of the input / output (which was rectified by reviewing the above-linked PR).

For posterity:
- `+cap` just checks whether the second-to-most significant bit of the input is `1` or `0`
- `+mas` just replaces the two most significant bits of the input with `01`
- `+peg` (as @eamsden explains above) just concatenates the bits of the input, excluding the most significant bit of either, and then sets the next most significant bit after the concatenation to `1`

ashelkovnykov:
@eamsden I applied your changes.

There was a "wrong bitshift direction" bug that didn't get caught by the tests, because it would allocate too many words, so there was no actual problem (just memory inefficiency). I added tests for the case, but it looks like the bug is not detectable after the fact because the excessive size is resolved by normalization to atom.

I'll leave the tests in, but I'm not sure how we would catch bugs like this with tests.

eamsden:
> @eamsden I applied your changes.
> 
> There was a "wrong bitshift direction" bug that didn't get caught by the tests, because it would allocate too many words, so there was no actual problem (just memory inefficiency). I added tests for the case, but it looks like the bug is not detectable after the fact because the excessive size is resolved by normalization to atom.
> 
> I'll leave the tests in, but I'm not sure how we would catch bugs like this with tests.

Good catch!

The test would have to be somewhat imprecise and check that the allocation didn't exceed what was reasonable. That wrong bitshift would have increased the allocation size from what was necessary by a factor of 4k, so it shouldn't be too hard to catch with thresholding. I'll make an issue to have tests like this.

rust/ares/src/jets/tree.rs
```diff
@@ -64,16 +63,15 @@ pub fn jet_peg(context: &mut Context, subject: Noun) -> Result {
     let b_bits = met(0, b);
     let out_bits = a_bits + b_bits - 1;
 
-    let out_words = (out_bits + 63) << 6; // bits to 8-byte words
+    let out_words = (out_bits + 63) >> 6; // bits to 8-byte words
```
eamsden, line 67: good catch!

eamsden (approved):

PR #118, opened 2023-10-27
Jets for `+mure`, `+mute`, `+mule`, and `+mole`

ashelkovnykov:
Note that these jets are run in a child road in Vere, but not in Ares (we shortcut the call to `interpret()` without using `mink()` or `jet_mink()`). We handle the scry handler stack ourselves in these jets, so to make the road transition official, all we would have to do is perform the computation with a fresh cache. Not sure if that's necessary, though.

### DIFF
--- a/rust/ares/src/jets/nock.rs
+++ b/rust/ares/src/jets/nock.rs
@@ -4,7 +4,7 @@ use crate::hamt::Hamt;
 use crate::interpreter::{interpret, Context, Error};
 use crate::jets::util::slot;
 use crate::jets::{JetErr, Result};
-use crate::noun::{Noun, D, T};
+use crate::noun::{Cell, Noun, D, NO, T, YES};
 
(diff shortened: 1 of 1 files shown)


rust/ares/src/jets/nock.rs
```diff
@@ -91,6 +149,32 @@ pub mod util {
     pub const LEAF: Noun = D(tas!(b"leaf"));
     pub const ROSE: Noun = D(tas!(b"rose"));
 
+    /// The classic "slam gate" formula.
```
eamsden, line 152: Just a marker for me: we should have a macro to create literal nouns as byteslices so we can just write them out as brackets, bytestrings, and integer literals, and so they can live in the data segment of the executable and not have to be allocated on the stack everywhere.

eamsden (approved):
LGTM

PR #119, opened 2023-10-27
Add jet for `+zing`

ashelkovnykov:
Add jet for `+zing` and use it to correctly weld the `$tang` of trace messages prior to passing the trace as input to `+mook`.


rust/ares/src/jets/text.rs
```diff
@@ -35,6 +45,54 @@ pub mod util {
         }
         Ok(len)
     }
+
+    pub fn zing(stack: &mut NockStack, mut list: Noun) -> jets::Result {
```
eamsden, line 49: This would be *much* clearer if written in destination-passing / "tail modulo cons" style.

- Set `res = D(0)`
- Keep a mutable noun pointer, initially `&mut res`.
- Whenever a new list element is discovered, allocate a new cell for it. Write the cell through the destination pointer, write the element to the head of the cell, set the pointer to the tail as the new destination pointer.
- When the outer loop exits, write `D(0)` to the destination pointer, then return `res`.

eamsden (changes requested):

rust/ares/src/jets/text.rs
eamsden: I made this change in 4500427e4ae28031cac79b4fd13ba3e17311ff3b / `eamsden/zing`

ashelkovnykov: Yes, this is clearly superiour

eamsden (approved):

PR #131, opened 2023-11-21
Use Nix shell in CI

eamsden:


eamsden:
@ashelkovnykov PTAL. Note that this will require a bit of additional work in the urcrypt PR as well

.github/workflows/ares-shared.yml
```diff
@@ -3,40 +3,50 @@ on:
...
+      - uses: cachix/install-nix-action@v23
+        name: Install Nix
+        with:
+          extra_nix_config: "extra-experimental-features = nix-command flakes"
 
+      # Check formatting
       - name: Format
-        run: cargo fmt --check
+        run: |
+          cd ares
```
ashelkovnykov, line 27: Why do we set the working directory to `rust` and then manually `cd` into the `ares` dir for each command?

rust/shell.nix
ashelkovnykov, line 1: Why is `shell.nix` removed?

ashelkovnykov (commented):
@eamsden I took a look, but I don't know Nix, so aside from looking at the Git workflow YAML, I just eye-balled it for "looks correct-ish". @matthew-levan Can you take a look too?

rust/shell.nix
eamsden: We're going to use `flake.nix` and `nix develop` instead of `shell.nix` and `nix-shell`. It's not worth fighting the ecosystem which has almost entirely moved to flakes.

rust/flake.nix
```diff
@@ -0,0 +1,32 @@
...
+          buildInputs = [
+            (fenix.packages.${system}.complete.withComponents [
+              "cargo"
+              "clippy"
+              "rustc"
+              "rustfmt"
+              "rust-src"
+            ])
+            pkgs.cargo-watch
+            pkgs.gdb
```
matthew-levan, line 26: Should we add the `urcrypt` dependencies here too?

ashelkovnykov (approved):

PR #132, opened 2023-11-21
naive codegen interpreter

matthew-levan:


eamsden:
@matthew-levan work on this should proceed from the head of #135

PR #134, opened 2023-11-24
Load codegen from included jam

eamsden:


eamsden:
Subsuming this in #135

PR #135, opened 2023-11-24
Codegen for Ares

eamsden:
Supersedes #89 with much cleaner history and rebased onto status. @matthew-levan work on integrating the codegen into rust-side Ares should happen here.

**Edward TODO**
- [x] Fix missing cases in linearizer
- [ ] cons/decons and move renaming in linearizer
- [ ] `%phi` and `%hip` elimination in linearizer (minimize register set?)
- [x] move `+wink` to a separate hoon file or even generator (make sure axes are re-established in rust-side code)
- [x] tag top-level codegen core with an incrementing version constant, and check it on rust side (make sure axes are re-established in rust-side code)
- [x] In-depth review of rust-side execution code
- [ ] Thoroughly review linearizer crash and poison logic
- [ ] Thread "which interpreter" enum through context
- [ ] Thread slow stack through jets and jet helpers
- [ ] Add to or delete from `hill` hamt by diffs emitted from `+poke` in codegen core
- [ ] Make jets and helpers use "which interpreter" enum


rust/ares/src/main.rs
```diff
@@ -80,8 +82,11 @@ fn main() -> io::Result<()> {
         hot,
         scry_stack: D(0),
         trace_info: None,
-        line: D(0),
+        line: None,
```
eamsden, line 83: This is the only change that should be made to `main.rs()` prior to a refactor making it just a thin wrapper around `serf::serf()`

eamsden: @matthew-levan

rust/ares/src/codegen.rs
```diff
@@ -494,6 +518,71 @@ pub fn cg_interpret(context: &mut Context, subject: Noun, formula: Noun) -> Resu
...
+                        loop {
+                            if v.raw_equals(D(0)) {
+                                if !walt.raw_equals(D(0)) {
+                                    Err(Error::NonDeterministic(D(0)))?;
+                                };
+                                break;
+                            };
+
+                            let v_i = slot(v, 2)?.as_direct()?.data();
+                            v = slot(a, 3)?;
```
joemfb, line 579: This should be `v = slot(v, 3)?;`

eamsden: Yup

rust/ares/src/codegen.rs
```diff
@@ -716,11 +716,78 @@ fn cg_interpret_inner(
...
+                    let p = slot(bend, 14).unwrap().as_direct().unwrap().data() as usize;
+                    let d = slot(bend, 30).unwrap().as_direct().unwrap().data() as usize;
+                    let mut t = slot(bend, 31).unwrap();
+
+                    if let Some(cell) = context.scry_stack.cell() {
+                        let scry_ref = register_get(current_frame, e);
+                        let scry_path = register_get(current_frame, p);
+                        let scry_stack = context.scry_stack;
+                        let scry_handler = cell.head();
+                        let scry_gate = scry_handler.as_cell().unwrap();
```
eamsden, line 730: Scry stack deconstruction should nock-crash on failure. @matthew-levan

rust/ares/src/codegen.rs
```diff
@@ -0,0 +1,1294 @@
...
+                }
+                tas!(b"inc") => {
+                    let s = slot(pole, 6).unwrap().as_direct().unwrap().data() as usize;
+                    let d = slot(pole, 7).unwrap().as_direct().unwrap().data() as usize;
+                    let s_value = register_get(current_frame, s);
+                    if let Ok(atom) = s_value.as_atom() {
+                        let value = inc(&mut context.stack, atom);
+                        register_set(current_frame, d, value.as_noun());
+                    } else {
+                        ActualError(Error::Deterministic(D(0)));
```
eamsden, line 237: This expression is evaluated *and then thrown away by the `;`*. It will not stop the loop. 
```suggestion
                        break Err(ActualError(Error::Deterministic(D(0))));
```

rust/ares/src/codegen.rs
```diff
@@ -0,0 +1,1294 @@
...
+                    if let Some(cell) = context.scry_stack.cell() {
+                        let scry_ref = register_get(current_frame, e);
+                        let scry_path = register_get(current_frame, p);
+                        let scry_stack = context.scry_stack;
+                        let scry_handler = cell.head();
+                        let scry_gate;
+                        if let Ok(cell) = scry_handler.as_cell() {
+                            scry_gate = cell;
+                        } else {
+                            return Err(ActualError(Error::ScryCrashed(D(0))));
```
eamsden, line 715: We shouldn't ever `return` directly from the loop. Just break from it. This allows easier debugging because we can insert debugging code after the loop exit
```suggestion
                            break Err(ActualError(Error::ScryCrashed(D(0))));
```

eamsden (commented):
@matthew-levan there are a lot of instances of missing break statements where errors should be returned but won't be. A couple of examples are below but the error is pervasive. Please correct all instances.

PR #140, opened 2023-11-28
jets: fix `+mook` jet mismatches and simplify logic

ashelkovnykov:


rust/ares/src/interpreter.rs
```diff
@@ -1422,9 +1427,9 @@ mod hint {
                         }
                     }
                     Err(_) => {
-                        let stack = &mut context.stack;
-                        let tape = tape(stack, "serf: %hela: mook error");
-                        slog_leaf(stack, &mut context.newt, tape);
+                        // +mook should only ever bail if the input is not [%2 (list)]. Since we control the input
+                        // for this call exactly, there must exist a programming error in Ares if this occurs.
```
eamsden, line 1431: Will this still be true when mook is used as a jet though?

ashelkovnykov: Can you explain why you think that's relevant here? We're not using `mook` as a jet here; we're crafting the input to be `[%2 list]` on line `1398`. Since `mook` can only bail when the input does not have this form, a returned `Err` here implies a programming error on our part, and therefore we should panic.

eamsden: Disregard, I was not properly following the code logic.

eamsden (approved):

PR #142, opened 2023-11-28
nock: fix slot access bug

ashelkovnykov:


rust/ares/src/noun.rs
```diff
@@ -1261,7 +1261,10 @@ pub trait Slots: private::RawSlots {
      * Retrieve component Noun at axis given as Atom, or fail with descriptive error
      */
     fn slot_atom(&self, atom: Atom) -> Result<Noun> {
-        self.raw_slot(atom.as_bitslice())
+        match atom.as_either() {
+            Left(direct) => self.slot(direct.data()),
+            Right(indirect) => self.raw_slot(indirect.as_bitslice()),
+        }
```
joemfb, line 1267: Why don't direct atoms implement `as_bitslice()`?

ashelkovnykov: They do. I could was just waterfalling the call into the function defined above so that we check for `0`, since the slot code using bit slices assumes a non-zero value.

joemfb: Oh, I see.

eamsden: @ashelkovnykov since `raw_slot()` already returns a result, is there a reason not to just do the zero check there and not need to do it in any of the higher-level calls?

eamsden (approved):
This fix is correct. Please do reply to the comment about code structure but that is not a blocker on merging.

rust/ares/src/noun.rs
ashelkovnykov: @eamsden Right now it was the path of least resistance, but you're correct: we could just convert the `Option` to a `Result` and stick with only `BitSlice` implementations. I think this was one of the first things I touched when starting to work on Ares, and wasn't yet familiar with what is idiomatic in Rust.

PR #143, opened 2023-11-29
Integrate PMA

eamsden:
Integrate the PMA work done by @barter-simsum and add logic to stash nouns and cold state in it.

## goals for this PR
We intend to integrate this PR once the following are possible:
- [x] booting to a dojo prompt
- [x] not crashing on sync()
- [x] restoring the pier state and continuing with an interactive dojo prompt

Further issues will be opened for bugfixes and TODOs, but this will allow us to integrate codegen and other pending work against a stable interface.

## archived TODO

**Edward TODO**
- [x] High-level rust bindings
- [x] composable `Persist` trait
- [x] tie PMA into serf
- [x] strip out old PMA and double-jam code
- [x] Change `Hamt` to be pointer-to-stem instead of stem. (It should have been all along, and this facilitates `handle_from_u64`)
- [x] Implement `Persist` for `Noun`
- [x] Implement `Persist` for `Hamt` :sweat: 
- [x] Implement `Persist` for `Batteries`
- [x] Implement `Persist` for `BatteriesList`
- [x] Implement `Persist` for `NounList`
- [x] Implement `Persist` for `Cold`
- [x] Make unifying equality dirty things in the PMA

(after Alex's review)

- [x] Refactor save logic and move into `serf`
- [x] XX comments about where to put warm and hot state so they are not blown away by stack resets
- [x] Remove StemTraversalEntry
- [x] Fix incomplete and out-of-date comments
- [x] Make `Persist for Noun` use `Persist for Atom` implementation for atom case
- [x] Push cell children onto stack in `copy_to_buffer` for `Noun`
- [x] Remove unnecessary `continue`s in HAMT persist loops
- [ ] Make named pointer local vs handle destructuring consistent in `Persist` instances
- [x] Heavy-duty doc-comment on the implementation of the `Preserve` trait and the phase-separation between size-counting and actual copying.
- [x] Heavy-duty doc-comment on the representation of HAMTs

**David TODO**
- [x] add incremented txn id logic to `bt_sync`
- [x] write routine to restore memory maps (became obvious after implementing `_bt_data_cow`
- [x] check return of all mmap calls to not be `MAP_FAILED` - abort if so -- actually should check that the return is the 
fixed addr passed in since mmap makes no guarantee that it will map at that address, but we rely upon that
- [x] complete implementation of bt_state_close() - got neglected
- [ ] c test to verify integrity of btree on a `bt_state_close()` `bt_state_open()` cycle
- [ ] more thorough c tests to validate `_bt_insert` that will exercise deletion coalescing
- [x] `bt_malloc` test - validate modifications to mlist, flist, and possibly btree though that may be adequately covered by `_bt_insert`
- [ ] `bt_free` test. validate modifications to mlist and possibly btree
- [ ] possibly a test that will combine enough `bt_malloc` and `bt_free` calls and follow that up with a call to `bt_sync` to cover a scenario that will drop at least one tree
- [ ] remove CAN_COALESCE macro defn and usage. btree deletion coalescing implemented
- [ ] remove c-side warnings generated during `cargo build`

**Alex TODO**
- [ ] PMA code reviewed (C-side)
- [x] PMA integration code reviewed (Rust-side)


rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+
+impl<T: Copy + Persist> Persist for Hamt<T> {
+    unsafe fn space_needed(&mut self, stack: &mut NockStack, pma: &PMA) -> usize {
+        if pma.contains(self.0, 1) {
+            return 0;
+        }
+        let mut bytes: usize = size_of::<Stem<T>>();
+        if pma.contains((*self.0).buffer, (*self.0).size()) {
+            return bytes;
+        };
```
ashelkovnykov, line 592: When would this ever happen? If `buffer` is already in the PMA, that implies that it's unchanged from a previous event, and therefore the `bitmap` and `typemap` members should also be unchanged, since they're metadata derived from the `buffer`.

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+            traversal[depth].bitmap = traversal[depth].bitmap >> (next_chunk + 1);
+            traversal[depth].typemap = traversal[depth].typemap >> (next_chunk + 1);
+            traversal[depth].buffer = traversal[depth].buffer.add(1);
+
+            if next_type {
+                // true->stem false->leaf
+                // found another stem
+                traversal[depth + 1] = next_entry.stem;
+
+                if pma.contains(traversal[depth + 1].buffer, traversal[depth + 1].size()) {
```
ashelkovnykov, line 623: Shouldn't we check that `&next_entry` isn't already in the PMA first? And if it isn't, don't we need to add `size_of::<Stem<T>>()` to bytes?

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+                    continue;
+                }
+
+                // count the buffer for the next stem
+                bytes += traversal[depth + 1].size() * size_of::<Entry<T>>();
+                depth += 1;
+            } else {
+                let mut leaf = next_entry.leaf;
```
ashelkovnykov, line 631: Same as above comment, but for `Leaf`

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+            let next_chunk = traversal[depth].bitmap.trailing_zeros();
+            let next_type = traversal[depth].typemap & (1 << next_chunk) != 0;
+            let next_entry_ptr = traversal[depth].buffer;
+
+            traversal[depth].bitmap = traversal[depth].bitmap >> (next_chunk + 1);
+            traversal[depth].typemap = traversal[depth].typemap >> (next_chunk + 1);
+            traversal[depth].buffer = traversal[depth].buffer.add(1);
+
+            if next_type {
+                // Stem case
```
ashelkovnykov, line 700: There's no check here for whether the `Stem` is already in the PMA

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+                let stem_buffer_ptr = *buffer as *mut Entry<T>;
+
+                copy_nonoverlapping((*stem_ptr).buffer, stem_buffer_ptr, stem_size);
+                *buffer = stem_buffer_ptr.add(stem_size) as *mut u8;
+
+                (*stem_ptr).buffer = stem_buffer_ptr;
+
+                traversal[depth + 1] = *stem_ptr;
+                depth += 1;
+                continue;
```
ashelkovnykov, line 714: Unnecessary `continue`

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+                        .0
+                        .copy_to_buffer(stack, pma, buffer);
+                    (*(*leaf_ptr).buffer.add(leaf_idx))
+                        .1
+                        .copy_to_buffer(stack, pma, buffer);
+
+                    leaf_idx += 1;
+                }
+
+                continue;
```
ashelkovnykov, line 738: Unnecessary `continue`

rust/ares/src/jets/cold.rs
```diff
@@ -143,6 +200,55 @@ struct BatteriesListMem {
...
+                break;
+            }
+
+            let list_mem_ptr = *buffer as *mut BatteriesListMem;
+            copy_nonoverlapping((*dest).0, list_mem_ptr, 1);
+            *buffer = list_mem_ptr.add(1) as *mut u8;
+            (*dest).0 = list_mem_ptr;
+
+            (*(*dest).0).batteries.copy_to_buffer(stack, pma, buffer);
+            dest = &mut (*(*dest).0).next;
```
ashelkovnykov, line 239: There's a lot of inconsistency between using the named pointer or `(*dest).0` in the `Persist` implementations in this file. It would be clearer to stick with the named pointer, e.g.:
```
(*dest).0 = list_mem_ptr;

(*list_mem_ptr).batteries.copy_to_buffer(stack, pma, buffer);
dest = &mut (*list_mem_ptr).next;
```

ashelkovnykov (commented):
This review is for the Rust side of the PMA integration code only. I'd recommend unit tests for `space_needed` and `copy_to_buffer` for  every single class that implements the `Persist` trait.

rust/ares/src/jets/cold.rs
eamsden: Noted, good catch

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    #[inline]
+    fn meta_set(&self, field: BTMetaField, val: u64) {
+        unsafe { bt_meta_set(self.0, field as usize, val) };
+    }
+
+    pub unsafe fn contains<T>(&self, ptr: *const T, count: usize) -> bool {
+        bt_inbounds(self.0, ptr as *mut c_void) != 0
+            && bt_inbounds(self.0, ptr.add(count) as *mut c_void) != 0
```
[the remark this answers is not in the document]
eamsden: > Why...check the end of the allocation as well?
Defensive programming, this is persistence code. Though, if T is in we should simply assert T+count is also in.

> Why.. "in the possible bounds of the PMA"?
Perhaps in contradiction to the above, but there's not an easy way to get the furthest-allocated virtual-memory boundary of the PMA.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+
+pub trait Persist {
+    /// Count how much space is needed, in bytes. May set marks so long as marks are cleaned up by
+    /// [copy_into_buffer]
+    unsafe fn space_needed(&mut self, stack: &mut NockStack, pma: &PMA) -> usize;
+
+    /// Copy into the provided buffer, which may be assumed to be at least as large as the size
+    /// returned by [space_needed] on the same structure. Return a u64 handle that could be saved
+    /// in metadata
+    unsafe fn copy_to_buffer(&mut self, stack: &mut NockStack, pma: &PMA, buffer: &mut *mut u8);
```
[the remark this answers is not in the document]
eamsden: Comment is out of date. Good catch.

Modifying self is consistent with what `Preserve` instances do, and isn't really tricky given that basically all instances of this type are newtype wrappers around pointers. So you're just copying something and then modifying the pointer you were passed in situ.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    /// Persist an object into the PMA using [space_needed] and [copy_to_buffer], returning
+    /// a [u64] (probably a pointer or tagged pointer) that can be saved into
```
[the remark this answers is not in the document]
eamsden: good catch

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/// Unmark an allocated noun
+unsafe fn unmark(a: Allocated) {
+    let metadata = a.get_metadata();
+    a.set_metadata(metadata & !NOUN_MARKED);
+}
+
+impl Persist for Atom {
+    unsafe fn space_needed(&mut self, stack: &mut NockStack, pma: &PMA) -> usize {
```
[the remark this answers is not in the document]
eamsden: No.

The entire reason for splitting this into two phases is that each phase composes individually. This should almost certainly be called out in detail in a doc comment on the `Persist` trait.

`space_needed` on the instance of a containing type should only ever call `space_needed` on a contained type, never `copy_to_buffer`. Similarly, `copy_to_buffer` should only ever call `copy_to_buffer`, not `space_needed`. And neither should ever call `save_to_pma`.

The whole idea is that you iterate `space_needed` over your entire e.g. `Snapshot` first, and then allocate one buffer in the PMA, then call `copy_to_buffer` over your entire `Snapshot`.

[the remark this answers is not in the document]
eamsden: In fact, I can't think of a reason `save_to_pma` should ever be re-implemented from the default. So it should probably live in `impl PMA` with a trait-bounded generic, not in the trait itself.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+        if let Ok(indirect) = self.as_indirect() {
+            let count = indirect.raw_size();
+            if !pma.contains(indirect.to_raw_pointer(), count) {
+                if !mark(indirect.as_allocated()) {
```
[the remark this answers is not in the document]
eamsden: The definition of `mark` is that it marks if not already marked, and then returns whether the `Allocated` was already marked. Marking is done by setting the high bit of metadata. There should be no forwarding pointers at any point during a `space_needed` call. In fact, an assertion to this effect is probably in order, this being persistence code.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+                if let Some(forward) = indirect.forwarding_pointer() {
+                    *self = forward.as_atom();
+                } else {
+                    let indirect_buffer_ptr = *buffer as *mut u64;
+                    copy_nonoverlapping(indirect.to_raw_pointer(), indirect_buffer_ptr, count);
+                    *buffer = indirect_buffer_ptr.add(count) as *mut u8;
+
+                    indirect.set_forwarding_pointer(indirect_buffer_ptr);
+
+                    *self = IndirectAtom::from_raw_pointer(indirect_buffer_ptr).as_atom();
```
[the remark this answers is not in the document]
eamsden: no, self is a mutable reference to the tagged indirect pointer. indirect is the original tagged indirect pointer, and is invalidated once the forwarding pointer is set. Comment forthcoming

rust/ares/src/hamt.rs
```diff
@@ -161,10 +163,11 @@ impl<T: Copy> MutHamt<T> {
 }
 
 #[repr(packed)]
+#[repr(C)]
 struct Stem<T: Copy> {
     bitmap: u32,
     typemap: u32,
-    buffer: *const Entry<T>,
```
ashelkovnykov, line 167: I don't think this should be necessary - change to the buffer implies a new `Stem`, doesn't it?

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+                        if !pma.contains(indirect.to_raw_pointer(), count) {
+                            if !mark(allocated) {
+                                space += count * size_of::<u64>();
+                            }
+                        }
+                    }
+                    Right(cell) => {
+                        if !pma.contains(cell.to_raw_pointer(), 1) {
+                            if !mark(allocated) {
+                                space += size_of::<CellMemory>();
```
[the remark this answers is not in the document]
eamsden: :scream: good catch

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+
+            match noun.as_either_direct_allocated() {
+                Left(direct) => {
+                    *dest = noun;
+                }
+                Right(allocated) => {
+                    if let Some(a) = allocated.forwarding_pointer() {
+                        *dest = a.as_noun();
+                        continue;
+                    }
```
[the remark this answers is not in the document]
eamsden: Why would that happen? The only other place forwarding pointers are set is in the noun `Preserve` instance and no live noun should ever contain a forwarding pointer if we're not in the middle of a `preserve` or `copy_to_buffer` call. If the former we're not going to overlap with `copy_to_buffer` and if the latter then the forwarding pointer precisely means that we already copied this noun to the PMA.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+
+                    match allocated.as_either() {
+                        Left(mut indirect) => {
+                            let count = indirect.raw_size();
+                            if pma.contains(indirect.to_raw_pointer(), count) {
+                                *dest = noun;
+                                continue;
+                            }
+
+                            unmark(allocated);
```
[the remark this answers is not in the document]
eamsden: "unmark" is clearing the mark bit in the metadata set by the sizing call, so it does not pollute the noun's metadata in the PMA. Every cell and indirect atom that we actually copy into the PMA will have been marked by `space_needed` and will need to be `unmarked` though @frodwith suggested an optimization to avoid the latter step last CCRS.

rust/ares/src/serf.rs
```diff
@@ -82,8 +89,41 @@ impl Context {
...
+        unsafe {
+            self.arvo = (*(snapshot.0)).arvo;
+            self.nock_context.cold = (*(snapshot.0)).cold;
+        }
+
+        // Reset the nock stack, freeing all memory used to compute the event
+        self.nock_context.stack.reset(0);
+
+        // Since we reset the nockstack the stack-allocated hot state isn't valid anymore
+        self.nock_context.hot = Hot::init(&mut self.nock_context.stack);
```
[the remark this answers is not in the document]
eamsden: Yes there should be an `XX` comment here about being able to save the state of the stack and reset to that, thus preserving e.g. the hot state. Also, because allocating the hot state HAMT will generate garbage, we should push, generate the hot state, and copy it into the top frame to keep it there.

But neither of these are expedient now.

rust/ares_pma/build.rs
```diff
@@ -0,0 +1,83 @@
+extern crate bindgen;
+
+use std::env;
+use std::path::PathBuf;
+
+use bindgen::CargoCallbacks;
+
+fn main() {
+    let profile = env::var("PROFILE").unwrap();
+    let opt_level = match profile.as_ref() {
```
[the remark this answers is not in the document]
eamsden: It should be cc @barter-simsum

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
-                let (stem, mut position) = traversal_stack[traversal_depth - 1]
-                    .expect("Attempted to access uninitialized array element");
-                // can we loop over the size and count leading 0s remaining in the bitmap?
-                'preserve_stem: loop {
-                    if position >= 32 {
-                        traversal_depth -= 1;
-                        continue 'preserve;
+        if stack.is_in_frame(self.0) {
+            let dest_stem = stack.struct_alloc_in_previous_frame(1);
+            copy_nonoverlapping(self.0, dest_stem, 1);
```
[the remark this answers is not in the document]
eamsden: copy_nonoverlapping() and rust memory function general take counts in units of the size of the pointed-to type, not bytes. So since our pointer is to Stem, the `1` is "1 `Stem`" not "1 byte"

rust/ares/src/serf.rs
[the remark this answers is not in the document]
eamsden: The above is done

rust/ares/src/serf.rs
```diff
@@ -31,26 +30,34 @@ const FLAG_TRACE: u32 = 1 << 8;
...
-    pub fn new(snap_path: &PathBuf, trace_info: Option<TraceInfo>) -> Self {
+    pub fn new(snap_path: PathBuf, trace_info: Option<TraceInfo>) -> Self {
         // TODO: switch to Pma when ready
         // let snap = &mut snapshot::pma::Pma::new(snap_path);
-        let mut snapshot = DoubleJam::new(snap_path);
         let mut stack = NockStack::new(1024 << 10 << 10, 0);
         let newt = Newt::new();
-        let cache = Hamt::<Noun>::new();
+        let cache = Hamt::<Noun>::new(&mut stack);
+        let pma = PMA::open(snap_path).unwrap();
```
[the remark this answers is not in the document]
eamsden: I did a big refactor on this code, please re-review.

rust/ares/src/serf.rs
```diff
@@ -82,8 +89,41 @@ impl Context {
...
+            (*snapshot_mem_ptr).cold = self.nock_context.cold;
+            let mut snapshot = Snapshot(snapshot_mem_ptr);
+            self.pma.save(&mut self.nock_context.stack, &mut snapshot);
+            snapshot
+        };
+
+        // reset pointers in context to PMA
+        unsafe {
+            self.arvo = (*(snapshot.0)).arvo;
+            self.nock_context.cold = (*(snapshot.0)).cold;
```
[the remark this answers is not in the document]
eamsden: ibid

rust/ares/src/serf.rs
```diff
@@ -82,8 +89,41 @@ impl Context {
...
-        self.snapshot
-            .save(&mut self.nock_context.stack, &mut self.arvo);
+        let snapshot = unsafe {
+            let snapshot_mem_ptr: *mut SnapshotMem = self.nock_context.stack.struct_alloc(1);
+
+            // Save into PMA (does not sync)
+            (*snapshot_mem_ptr).epoch = self.epoch;
+            (*snapshot_mem_ptr).event_num = self.event_num;
+            (*snapshot_mem_ptr).arvo = self.arvo;
+            (*snapshot_mem_ptr).cold = self.nock_context.cold;
```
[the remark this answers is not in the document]
eamsden: ibid

rust/ares/src/hamt.rs
```diff
@@ -481,82 +486,265 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
                         }
                     }
                 }
             }
         }
     }
 }
+
+#[derive(Copy, Clone)]
+struct StemTraversalEntry<T: Copy> {
```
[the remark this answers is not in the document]
eamsden: I think that's correct

rust/ares/src/hamt.rs
eamsden: `next_entry` is temporarily copied from the buffer we are traversing. `Entry` is a union, so the `.stem` here is a union, not struct, field. Entries, not pointers to entries, are elements of the buffers.

rust/ares/src/hamt.rs
eamsden: Same reply: `Leaf` lives in a `Stem`'s buffer as an `Entry`

rust/ares/src/hamt.rs
eamsden: No, the allocated objects are just the buffers.

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+                return self.handle_to_u64();
+            }
+
+            let space_as_pages = (space + (BT_PAGESIZE as usize - 1)) >> BT_PAGEBITS;
+
+            let mut buffer = bt_malloc(pma.0, space_as_pages) as *mut u8;
+            let orig_buffer = buffer;
+            self.copy_to_buffer(stack, pma, &mut buffer);
+            assert!(orig_buffer.offset_from(buffer) > 0);
+            assert!(orig_buffer.offset_from(buffer) <= space.try_into().unwrap());
```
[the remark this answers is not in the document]
eamsden: > always `=`
No, `<` is not an unallocated write, `>` would be.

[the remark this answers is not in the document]
eamsden: `space.into()` is probably what should be done here. The point is to have it checked as we are going from an unsigned to signed word of the same bitwidth.

rust/ares/src/hamt.rs
eamsden: The structure of the HAMT is that metadata for the buffer is stored in the *parent'*s buffer, except at the top level where we allocate just one `Stem` struct. A parent can be changed when children aren't, so an altered buffer on the stack could reference an unaltered buffer in the PMA.

rust/ares/src/hamt.rs
eamsden: We copy_nonoverlapping the existing buffer first and then `copy_to_buffer` each stem with it, so the mutable pointer is necessary to enable this pattern. I think it is also necessary for unifying equality on keys during lookups

eamsden:
@ashelkovnykov I believe I've addressed all of your review comments except for the consistency of named pointer handles, which requires some extra consideration on my part that I'm going to push off for a bit.

eamsden:
@barter-simsum when you get a chance could you clean up the C-side warnings that are generated when build.rs builds btree.c ?

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,341 @@
...
+    unsafe fn handle_from_u64(meta_handle: u64) -> Self;
+}
+
+impl Persist for Snapshot {
+    unsafe fn space_needed(&mut self, stack: &mut NockStack, pma: &PMA) -> usize {
+        let mut arvo = (*(self.0)).arvo;
+        let mut cold = (*(self.0)).cold;
+        let arvo_space_needed = arvo.space_needed(stack, pma);
+        let cold_space_needed = cold.space_needed(stack, pma);
+        (((size_of::<SnapshotMem>() + 7) >> 3) << 3) + arvo_space_needed + cold_space_needed
```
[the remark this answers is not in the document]
ashelkovnykov: If:
```
#define MASK 0xfffffff8
```
then:
```
(((x + 7) >> 3) << 3) == (x + 7) & MASK
```
Is there a reason to prefer the former over the latter?

eamsden:
Notes from attempts to boot a fakedev with the PMA today:
- I fixed some silly errors in `Persist` and `Preserve` instances.
- I fixed the assertions in `bt_meta_set` and `bt_meta_get`
- It doesn't appear that `ftruncate()` or `truncate()` are actually called anywhere.
- It doesn't appear that `bt_malloc` results in the allocated range being mapped to the file offset returned from the freelist. The first attempt to `save_to_pma` throws SIGBUS from the first copy. Adding `ftruncate` does not fix this, and examining `/proc/<PID>/maps` shows only a whole-PMA linear mapping to the file.

@barter-simsum we should use the pairing session tomorrow to try to get to boot/shutdown/restore for a fakedev.

eamsden:
Things to verify
- [ ] `ftruncate()` is properly used to set/extend data.pma
- [ ] `bt_malloc()` properly `mmap`'s the allocated memory range to the allocated disk range
- [ ] `bt_sync()` properly sets protection on cleaned ranges to `PROT_READ` only
- [ ] `bt_free()` properly maps memory range with flags `MAP_FIXED | MAP_ANONYMOUS | MAP_NORESERVE` and protection `PROT_NONE`

@barter-simsum @ashelkovnykov

rust/ares/src/hamt.rs
ashelkovnykov: > A parent can be changed when children aren't, so an altered buffer on the stack could reference an unaltered buffer in the PMA

Correct, but not at the very top level. Maybe I'm missing and you can think of a counter example, but as I understand it an `insert` call on `Hamt` will always create a new buffer (if you look at the usages of `dest` in `insert`). Therefore, if `pma.contains(self.0, 1)` is `false`, then `pma.contains((*self.0).buffer, (*self.0).size())` is also necessarily `false`.

rust/ares/src/hamt.rs
```diff
@@ -481,82 +497,255 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
+            traversal[depth].buffer = traversal[depth].buffer.add(1);
+
+            if next_type {
+                // true->stem false->leaf
+                // found another stem
+                traversal[depth + 1] = next_entry.stem;
+
+                if pma.contains(traversal[depth + 1].buffer, traversal[depth + 1].size()) {
+                    continue;
+                }
```
ashelkovnykov, line 631: This is going to needlessly iterate through the buffer of a `Stem` that's already in the PMA. We should be able to skip this as such:
```
if pma.contains(next_entry.stem.buffer, next_entry.stem.size()) {
    continue;
}

depth += 1;
traversal[depth] = next_entry.stem;
// count the buffer for the next stem
bytes += traversal[depth].size() * size_of::<Entry<T>>();
```

rust/ares/src/hamt.rs
```diff
@@ -481,82 +497,255 @@ impl<T: Copy + Preserve> Preserve for Hamt<T> {
...
         }
     }
 }
+
+impl<T: Copy + Persist> Persist for Hamt<T> {
+    unsafe fn space_needed(&mut self, stack: &mut NockStack, pma: &PMA) -> usize {
+        if pma.contains(self.0, 1) {
+            return 0;
+        }
+        let mut bytes: usize = size_of::<Stem<T>>();
```
ashelkovnykov, line 594: I believe we're currently forgetting to add the size of the root `Stem`s buffer to `bytes`.

ashelkovnykov (dismissed):

rust/ares/src/persist.rs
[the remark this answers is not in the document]
eamsden: **Argument:**
`space_needed` could be conservative (i.e. request more than is necessary) without this being an error condition in the code. I believe it to be exact right now, but if it were to be conservative this should be supported.

**Counterargument:**
Insisting on an exact match would help catch unexpected behaviors, especially missing object copies.

Let me think about this.

rust/ares/src/serf.rs
```diff
@@ -28,47 +29,157 @@ crate::gdb!();
...
         constant_hot_state: &[HotEntry],
-    ) -> Self {
-        // TODO: switch to Pma when ready
-        // let snap = &mut snapshot::pma::Pma::new(snap_path);
-        let mut snapshot = DoubleJam::new(snap_path);
-        let mut stack = NockStack::new(512 << 10 << 10, 0);
+    ) -> Context {
+        let mut pma = PMA::open(snap_path).expect("serf: pma open failed");
+
+        let snapshot_version = pma.meta_get(BTMetaField::SnapshotVersion as usize);
```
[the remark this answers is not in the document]
eamsden: This is not the case. Roots is just a place to put handles to things stored in the PMA. But yes, the names should be consistent, which also requires changing the C API.

cc @barter-simsum

rust/ares/src/hamt.rs
eamsden: No. If the root stem is already in the PMA then we don't need to do any copying. Otherwise we initialize the `bytes` counter to the size of one `Stem`: the root stem.

rust/ares/src/hamt.rs
eamsden: Not so. This is checking whether the buffer of a *child* `Stem` to the stem at `depth` is already in the PMA, in which case we will skip preserving it.

ashelkovnykov: Correct, but if the buffer of the child `Stem` is already in the PMA, it stands to reason that the buffers of any children of *that* `Stem` are already in the PMA as well. However, we're going to traverse and check for each of those children, because we've increased `depth` by `1`.

rust/ares/src/hamt.rs
ashelkovnykov: But if the buffer of the root `Stem` isn't in the PMA, we never increase `bytes` by the size of the buffer.

eamsden: I just realized you were right about this and fixed it.

rust/ares/src/persist.rs
```diff
@@ -265,6 +265,7 @@ impl Persist for Noun {
             }
         }
         *buffer = buffer_u64 as *mut u8;
+        stack.frame_pop();
```
ashelkovnykov, line 268: How did we both miss this...

eamsden: Dunno but I caught it when `flip_top_frame`'s assertion that it was, in fact, in the top frame failed.

rust/ares/src/hamt.rs
eamsden: We haven't though. Look a few lines below. We `continue` in this `if` but increment depth below the `if`. We do write the child node into the traversal stack at `depth + 1` but that does not mutate `depth`.

rust/ares/src/serf.rs
```diff
@@ -79,22 +188,37 @@ impl Context {
...
+    ///
+    /// ## Safety
+    ///
+    /// Preserves nouns and jet states in context and then calls [flip_top_frame].
+    /// Other stack-allocated objects needing preservation should be preserved between
+    /// [event_update] and invocation of this function
+    pub unsafe fn preserve_event_update_leftovers(&mut self) {
+        let stack = &mut self.nock_context.stack;
+        stack.preserve(&mut self.arvo);
+        stack.preserve(&mut self.nock_context.cold);
```
ashelkovnykov, line 218: Why is preserving cold necessary - shouldn't it be in the PMA by this point? And any dangling forward pointers in other objects resolved by calling preserve on those objects?

rust/ares/src/mem.rs
```diff
@@ -51,6 +50,7 @@ pub struct NockStack {
     alloc_pointer: *mut u64,
     /** MMap which must be kept alive as long as this NockStack is */
     memory: MmapMut,
+    /** PMA from which we will copy into the NockStack */
```
ashelkovnykov, line 53: Incoming member that's not yet present?

rust/ares/src/mem.rs
```diff
@@ -106,6 +143,26 @@ impl NockStack {
...
+    /** Start of the memory range for this NockStack */
+    pub fn get_start(&self) -> *const u64 {
+        self.start
+    }
+
+    /** End of the memory range for this NockStack */
+    pub fn get_size(&self) -> usize {
+        self.size
+    }
+
```
ashelkovnykov, line 165: I keep adding these for debugging, then removing them, too. Finally someone had the wherewithal to actually commit them...

rust/ares/src/persist.rs
```diff
@@ -0,0 +1,315 @@
...
+
+// round an address down to a page boundary
+fn bt_page_round_down<T>(ptr: *mut T) -> *mut c_void {
+    ((ptr as u64) & BT_PAGEBITS_MASK_OUT) as *mut c_void
+}
+
+// round an address up to a page boundary
+fn bt_page_round_up<T>(ptr: *mut T) -> *mut c_void {
+    (((ptr as u64) + (BT_PAGESIZE as u64) - 1) & BT_PAGEBITS_MASK_OUT) as *mut c_void
+}
```
ashelkovnykov, line 315: Anyway to have this in `btree.c` and call it from Rust? I don't like the idea of duplicating functions / constants between the two languages.

rust/ares/src/hamt.rs
ashelkovnykov: Bleh - you're right. Somehow I conflated "assignment to `traversal[depth + 1]`" with "`depth` incremented". I think I got confused because `traversal` changes, even though `depth` doesn't, in this case.

rust/ares/src/mem.rs
eamsden: unifying equality moved to its own module, so we need public interfaces to this stuff.

rust/ares/src/persist.rs
eamsden: These functions are unused in `btree.c`, which exposes its interface in page offsets, not byte-level pointers.

eamsden: I agree with your opinion, but the functions are not duplicated (they are only rust-side) and the constants are imported via rust-bindgen from `btree.h`

rust/ares/src/serf.rs
eamsden: It's not necessary, these two lines can be removed. I was just cut-and-pasta-ing the stack flip from the bottom of the serf loop

eamsden:
@ashelkovnykov @joemfb the plan is to merge this once it is capable of
- [x] booting to a dojo prompt
- [x] not crashing on sync()
- [ ] restoring the pier state and continuing with an interactive dojo prompt

Please review with an eye toward this merge. Bugs, TODOs, and cleanup which are not on the path to merging the above functionality with a CI greencheck are to be filed as issues and worked on in separate PRs.

I believe the interface for this is sufficiently stabilized that merging it will enable me to focus on integrating other components, most notably codegen.

ashelkovnykov (approved):
I understand 100% of the Rust changes and 15% of the C changes - shipping a review to get this out of monolithic status.

PR #158, opened 2023-12-04
virtualization: properly restore cold/warm states, cache, and scry stack when catching an error

eamsden:
Fixed a bug caught by @ashelkovnykov 

@ashelkovnykov please confirm my understanding. In particular:

- [ ] `jet_mute` and `jet_mure` had the same bug as `mink`
- [ ] `jet_mute` and `jet_mure` additionally did not properly reset the scry stack in the context when catching an error.
- [ ] (not changed yet) `mink` further does not properly reset the scry stack in the context when catching an error

### DIFF
--- a/rust/ares/src/jets/nock.rs
+++ b/rust/ares/src/jets/nock.rs
@@ -1,6 +1,5 @@
 /** Virtualization jets
  */
-use crate::hamt::Hamt;
 use crate::interpreter::{interpret, Context, Error};
 use crate::jets::util::slot;
 use crate::jets::{JetErr, Result};
@@ -36,16 +35,12 @@ pub fn jet_mink(context: &mut Context, subject: Noun) -> Result {
     let v_formula = slot(arg, 5)?;
     let scry_handler = slot(arg, 3)?;
 
-    let old_cache = context.cache;
-    let old_scry_stack = context.scry_stack;
-
-    context.cache = Hamt::<Noun>::new();
-    context.scry_stack = T(&mut context.stack, &[scry_handler, old_scry_stack]);
+    let snapshot = util::snapshot_and_virtualize(context, scry_handler);
 
     match util::mink(context, v_subject, v_formula) {
         Ok(noun) => {
-            context.cache = old_cache;
-            context.scry_stack = old_scry_stack;
+            context.cache = snapshot.cache;
+            context.scry_stack = snapshot.scry_stack;
             Ok(noun)
         }
         Err(error) => match error {
@@ -64,7 +59,7 @@ pub fn jet_mink(context: &mut Context, subject: Noun) -> Result {
                 // Error::ScryCrashed, jet_mink() compares the two scry handler stack Nouns> If they
                 // are identical, jet_mink() bails with Error::Deterministic. Otherwise, it forwards
                 // the Error::ScryCrashed to the senior virtualization call.
-                if unsafe { context.scry_stack.raw_equals(old_scry_stack) } {
+                if unsafe { context.scry_stack.raw_equals(snapshot.scry_stack) } {
                     Err(JetErr::Fail(Error::Deterministic(trace)))
                 } else {
                     Err(JetErr::Fail(error))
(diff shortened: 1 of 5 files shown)


ashelkovnykov:
* `jet_mure` and `jet_mute` were also not saving `Cold`/`Warm` state correctly
* `jet_mure` and `jet_mute` did not correctly reset the scry stack on deterministic error & blocked scry
* `mink` **does** reset the scry stack correctly in all cases (via `jet_mink`)

rust/ares/src/jets/nock.rs
```diff
@@ -90,21 +85,26 @@ pub fn jet_mure(context: &mut Context, subject: Noun) -> Result {
...
-                context.restore(saved);
-                context.scry_stack = context.scry_stack.as_cell()?.tail();
+            Error::ScryBlocked(_) => {
+                context.scry_stack = snapshot.scry_stack;
+                Ok(D(0))
+            }
+            Error::Deterministic(_) => {
+                context.cold = snapshot.cold;
+                context.warm = snapshot.warm;
+                context.scry_stack = snapshot.scry_stack;
```
eamsden, line 107: Do we not need to reset the cache here?

rust/ares/src/jets/nock.rs
```diff
@@ -116,29 +116,29 @@ pub fn jet_mute(context: &mut Context, subject: Noun) -> Result {
...
             Ok(T(&mut context.stack, &[YES, res]))
         }
         Err(error) => match error {
             // Since we are using the pass-through scry handler, we know for a fact that a scry
             // crash must have come from a senior virtualization context.
             Error::NonDeterministic(_) | Error::ScryCrashed(_) => Err(JetErr::Fail(error)),
             Error::ScryBlocked(path) => {
-                context.scry_stack = context.scry_stack.as_cell()?.tail();
-                context.restore(saved);
+                context.scry_stack = snapshot.scry_stack;
```
eamsden, line 132: Do we not need to restore `cold`/`warm`/`cache` before `smyt`()? Won't it be running with invalidated versions of those pointers?

rust/ares/src/jets/nock.rs
```diff
@@ -116,29 +116,29 @@ pub fn jet_mute(context: &mut Context, subject: Noun) -> Result {
...
                 //      return [[%leaf "mute.hunk"] ~] if not
                 let bon = util::smyt(&mut context.stack, path)?;
                 Ok(T(&mut context.stack, &[NO, bon, D(0)]))
             }
             Error::Deterministic(trace) => {
-                context.scry_stack = context.scry_stack.as_cell()?.tail();
-                context.restore(saved);
+                context.cold = snapshot.cold;
+                context.warm = snapshot.warm;
+                context.scry_stack = snapshot.scry_stack;
```
eamsden, line 141: As above: do we not restore the old `cache` on failure?

rust/ares/src/serf.rs
```diff
@@ -339,6 +341,8 @@ fn soft(context: &mut Context, ovo: Noun, trace_name: Option<String>) -> Result<
         Ok(res) => Ok(res),
         Err(error) => match error {
             Error::Deterministic(trace) | Error::NonDeterministic(trace) => {
+                context.nock_context.cold = cold_snapshot;
+                context.nock_context.warm = warm_snapshot;
```
eamsden, line 345: `cache` should possibly be reset to a new Hamt here, if this method is only used in serf. Otherwise it should be saved and restored, right?

eamsden (commented):
Mostly just: "shouldn't we restore the saved cache after failure?" in a bunch of different places. I might not be following something...

eamsden:
Notes from discussion with @ashelkovnykov this morning

- `interpret()` should restore everything except the scry stack on error (since mink needs to check the scry stack as of the error to see if it should re-throw scry-crashed or convert to another error
- All virtualization functions should call `util::mink`

ashelkovnykov:
Follow-up to @eamsden 's comments above: I tried to elide several calls and unnecessary restorations in `jet_mink`, `jet_mure`, `jet_mute`, but it's making more problems than it's solving. It's resulting in the logic getting slightly more spread out, but also - much worse than that - it's making the logic of when we do/don't have a valid cold state / warm state / cache / scry stack hard to follow.

Therefore, we're going to put all of the logic (except scry stack) into `interpret` itself, and then force all possible virtualization through `util::mink`. Then we can verify all of the `Context` is in the correct state by checking only three locations: `jets::nock::util::mink`, `interpreter::interpret`, and `interpreter::exit`.

rust/ares/src/jets/nock.rs
ashelkovnykov: The logic is that the error is going to be propagated up to either a higher level `+mink` which will catch it and reset the cache, or to the root serf process - meaning that the cache is going to be reset anyways, as the event processing is over.

See comments below for planned changes.

rust/ares/src/jets/nock.rs
ashelkovnykov: `+smyt` just parses a `$path` to a `$tank`

rust/ares/src/jets/nock.rs
ashelkovnykov: Same as above; see comments below for planned changes.

rust/ares/src/serf.rs
ashelkovnykov: The cache should not be saved / restored here, since it's empty before going into this call and either an error or a success mean that it's about to be invalidated (since it's not a persistent cache). However, you're right that we need to reset it explicitly in the error case: the call to `goof` should have an empty cache.

See comments below for planned changes.

eamsden (commented):
This looks good to me now, but it needs to be tested booting a fakezod and someone not me or @ashelkovnykov needs to review it.

ashelkovnykov:
@eamsden Currently failing on fakezod boot

PR #159, opened 2023-12-05
main: just be a serf, panic if not a serf

eamsden:


eamsden:
> Code looks fine. What's your reasons for this?
> 
> The first one that occurs to me is maintaining one fewer code path. Are there others?

This enables #160 and makes the executable <> library interface cleaner

joemfb (approved):
this will do

drbeefsupreme (approved):

PR #160, opened 2023-12-05
Pass in hot state

eamsden:
Pass in the hot state as a parameter to `serf()`.

This enables distributing executables with other hot states by simply importing the ares crate, and passing in the additional hot state to the `serf()` call in the executable's `main()`.

### DIFF
--- a/rust/ares/src/jets/hot.rs
+++ b/rust/ares/src/jets/hot.rs
@@ -4,7 +4,7 @@ use ares_macros::tas;
 use either::Either::{self, Left, Right};
 use std::ptr::null_mut;
 
-// const A_50: Either<u64, (u64, u64)> = Right((b"a", 50));
+// const A_50: Either<u8, (u64, u64)> = Right((tas!(b"a"), 50));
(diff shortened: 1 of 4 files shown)


rust/ares/src/jets/hot.rs
```diff
@@ -57,8 +57,10 @@ const K_139: Either<&[u8], (u64, u64)> = Right((tas!(b"k"), 139));
 //     (&[A_50, Left(b"mink")], 1, jet_mink),
 // ];
 
+pub type HotEntry = (&'static [Either<&'static [u8], (u64, u64)>], u64, Jet);
```
ynx0, line 60: As an outsider, it would be nice to have a comment with field names.

joemfb (approved):
this will do

drbeefsupreme:
I'm working on testing this today, should have some feedback later.

ashelkovnykov (approved):

drbeefsupreme:
This was already merged, but I wanted to note that I have a working jetpack going through this code path.

PR #162, opened 2023-12-06
Pq/reset cache

philipquirk:
Reset the cache in work_swap before sending %crud event.


eamsden (approved):
This is correct. `interpret()` will reset the cold and warm state on failure as of #158 but will not reset the cache, which would lead to a use-after-free if the cache were ever accessed while processing a `%crud` event.

The cache *is* reset each time around the serf loop but `work_swap` is called directly when work or play fail.

Issue #165, opened 2023-12-07
CI: cache rust


eamsden:
We should have Github Actions cache our rust artifacts in CI. This would speed up CI time tremendously

eamsden:
Done in #174

PR #169, opened 2023-12-08
ci: cache rust dependencies

eamsden:
This ensures that the cargo cache is keyed on the nix hash of the development shell, so it will be discarded if we updated the development environment. Only dependencies are cached. Basic trials appear to show an approximately 50% reduction in CI runtime.

Partial implementation of #165


eamsden:
Superseded by #174

PR #174, opened 2023-12-10
CI: Cache nix and rust dependencies

eamsden:
Includes changes from and supersedes #169. This caches the nix build and cuts CI time to about 1 minute when cache is hit. Cache misses should only occur from devshell or cargo updates or changes. 

Full implementation of #165

### DIFF
--- a/.github/scripts/arch/get-runner-arch
+++ b/.github/scripts/arch/get-runner-arch
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# This could be improved by separately looking up OS and architecture
+# but it will do for now.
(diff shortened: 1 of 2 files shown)


.github/scripts/arch/get-runner-arch
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# This could be improved by separately looking up OS and architecture
+# but it will do for now.
+for kv in "X64-Linux,x86_64-linux" "X64-macOS,x86_64-darwin" "ARM64-macOS,aarch64-darwin"; do
+  KEY=${kv%,*};
+  VAL=${kv#*,};
```
ashelkovnykov, line 7: I've never seen this syntax for bash before, but I verified that it works

ashelkovnykov (approved):

PR #178, opened 2023-12-13
2stackz: gc top frame by flipping polarity (take 2)

matthew-levan:
Reverts urbit/ares#177

### DIFF
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -165,12 +165,10 @@ impl NockStack {
             } else {
                 ptr_u64 >= self.alloc_pointer && ptr_u64 < prev
             }
+        } else if prev.is_null() {
+            ptr_u64 >= self.start && ptr_u64 < self.alloc_pointer
         } else {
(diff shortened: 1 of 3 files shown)


rust/ares/src/mem.rs
```diff
@@ -123,7 +160,13 @@ impl NockStack {
         let ptr_u64 = ptr as *const u64;
         let prev = *self.prev_stack_pointer_pointer();
         if self.is_west() {
-            ptr_u64 >= self.alloc_pointer && ptr_u64 < prev
+            if prev.is_null() {
```
ashelkovnykov, line 163: Please add comment documenting that these `null` checks are a necessity of the invariant that the top frame has a `null` `prev_stack_pointer`. This invariant is now responsible for changes to code to function correctly in two locations (`is_in_frame` and `unifying_equality`).

eamsden: done

ashelkovnykov (approved):

PR #182, opened 2023-12-14
Implement `bail: %meme`

### LINKED ISSUE #172: OOM / bail:meme check in NockStack allocation routines

ashelkovnykov:
Resolves #172 

* Adds OOM checks on `NockStack` allocations
* Implements a mote system

Current issues:
- [ ] Allocation issues; see [comment 1](https://github.com/urbit/ares/pull/182#issuecomment-1859205099) and [comment 2](https://github.com/urbit/ares/pull/182#issuecomment-1862077623) below
- [x] Need to redo some name changes in arguments to `interpret` which were made as a result of the way in which the `hw_exception::catch` closure can use local arguments (it can't use any `&mut` due to extending `UnwindSafe`)


rust/ares/src/serf.rs
```diff
@@ -185,6 +186,11 @@ pub fn serf(constant_hot_state: &[HotEntry]) -> io::Result<()> {
     signal_hook::flag::register_conditional_shutdown(SIGINT, 1, Arc::clone(&TERMINATOR))?;
     signal_hook::flag::register(SIGINT, Arc::clone(&TERMINATOR))?;
 
+    unsafe {
+        // Register a hook for SIGSEGV, which captures and throws a backtrace.
+        hw_exception::register_hook(&[hw_exception::Signo::SIGSEGV], |e| hw_exception::throw(e));
+    }
+
```
eamsden, line 193: What happens for a SIGSEGV outside the interpret closure passed to `hw_exception::catch` ?

rust/ares/src/interpreter.rs
```diff
@@ -389,537 +390,568 @@ pub fn interpret(context: &mut Context, mut subject: Noun, formula: Noun) -> Res
     //
     // (See https://docs.rs/assert_no_alloc/latest/assert_no_alloc/#advanced-use)
     let nock = assert_no_alloc(|| unsafe {
-        push_formula(&mut context.stack, formula, true)?;
+        hw_exception::catch(|| {
```
eamsden, line 403: is it safe to nest these invocations? What happens if a SIGSEGV results inside +mink or a parser jet?

ashelkovnykov: Yes, I believe nesting invocations is safe. It doesn't explicitly say so in the docs, but my understanding is that `catch` closures form a stack of `setjmp` buffers, and `throw` will `longjmp` to the one on the top of the stack. I've tested this experimentally and my results line up with this interpretation.

A `SIGSEGV` in `+mink` or one of the parser jets will be caught by the top-most `catch` closure.

rust/ares/src/serf.rs
ashelkovnykov: If there is another `catch` closure wrapping it, then that closure will catch it. Otherwise, `throw` will detect that there is no wrapping `catch` closure around the error and the process will terminate with an uncaught `SIGSEGV`.

ashelkovnykov:
Note that we now have two different crates registering signal hooks. Both `SIGINT` and `SIGSEGV` are being handled correctly, but it's something to keep in mind.

ashelkovnykov:
@eamsden `throw` appears to allocate internally, but I would assume it does so in a way that's acceptable to us, since it uses `longjmp` and no unwinding occurs. Might this be a rare case where we add a comment, wrap in `permit_alloc`, and carry on?

ashelkovnykov:
@eamsden On the back of the previous comment, I'm going to play around with using `hw_exception` for `SIGINT` as well, as discussed in #152

eamsden:
@ashelkovnykov the point of #152 is not to catch SIGINT with `hw_exception`, as SIGINT is asynchronous. Rather it is to turn `SIGINT` into `SIGSEGV` by mprotecting the entire NockStack, and then using `hw_exception` to catch the resulting SIGSEGV.

rust/ares/src/serf.rs
eamsden: Perfect.

rust/ares/src/interpreter.rs
eamsden: By "topmost" you mean "most inner" so the most recently-called `+mink` would be the one to receive the error, right?

eamsden:
@ashelkovnykov is this PR ready to merge? Or are you planning to also address #152 in the same PR?

rust/ares/src/interpreter.rs
ashelkovnykov: Yes; "top-most catch" `==` "inner-most `+mink`"

ashelkovnykov:
@eamsden I want to @eamsden I want to quickly see if I can improve the PR a little and unify how we handle interrupts + OOM

eamsden:
@ashelkovnykov as discussed can you mark this ready-for-review and work on #152 in a separate PR?

ashelkovnykov:
The nested `catch` statements are creating an issue due to `hw_exception::run_hooks` allocating internally (this function gets called when an exception is thrown to find the signal hook handler we register in `serf.rs`). Making this `assert_no_alloc` error go away requires nesting `permit_alloc` calls around each `hw_exception::catch`, which is miserable.

But it also doesn't fully resolve the issue: the `longjmp` inside of `throw` is messing up the internal state management of `assert_no_alloc`, which uses thread_local counters to count nested `assert_no_alloc` calls. Therefore, when you `longjmp` out, the code assumes that you're still inside of `assert_no_alloc` and triggers an error on `Newt` allocations.

To work around this, @eamsden is hacking together a Frankenstein's monster crate from both `hw_exception` and `assert_no_alloc`. While vendoring two crates into one isn't ideal, they're luckily both very small and `assert_no_alloc` in particular is basically frozen.

eamsden:
> To work around this, @eamsden is hacking together a Frankenstein's monster crate from both `hw_exception` and `assert_no_alloc`. While vendoring two crates into one isn't ideal, they're luckily both very small and `assert_no_alloc` in particular is basically frozen.

Thinking further this isn't necessary. If we make hw_exception not allocate, and nest `catch` inside every `assert_no_alloc` (we really will only have 2, one in `interpret` and one in `cg_interpret` when codegen merges) then no longjmp from `hw_exception::catch` should ever jump outside of an `assert_no_alloc` call.

eamsden:
looking even further: hw_exception is overengineered for purpose and unavoidably allocates when `catch`-ing an exception. Not sure what to do from here.

eamsden:
I've come to the conclusion that we need a custom crate that provides unified functionality of both `assert_no_alloc` and `hw_exception`, but with the latter specialized to SIGSEGV handling.

ashelkovnykov:
@eamsden That sounds right - and luckily there's plenty of prior art

ashelkovnykov:
Archived this PR branch [here](https://github.com/ashelkovnykov/ares/tree/guardpage); going to submit a new PR that just adds motes. Guard page work will now be done in collaboration w/ @matthew-levan in #202.

PR #183, opened 2023-12-15
main: add stop_for_debug feature to stop Ares when it starts for a debugger to attach

eamsden:


matthew-levan (approved):

matthew-levan:
DevEx score +420

PR #188, opened 2023-12-17
serf: implement %live %exit

### LINKED ISSUE #187: Serf: actually exit
Right now when the king sends us a `%live %exit` command we `eprintln("exit")`...and do not exit. We should simply exit at this point.

ashelkovnykov:
Resolves #187


eamsden (approved):
:shipit:

PR #189, opened 2023-12-17
`ares_crypto`

### LINKED ISSUE #156: Use rust crypto for crypto jets
In aid of #155, we should see if we can replace Urcrypt with the work of the [Rust Crypto team](https://github.com/RustCrypto).

matthew-levan:
This PR removes our dependency on `urcrypt` for binary builds by replacing the `urcrypt` functions we used for our currently-implemented crypto jets (`aes_siv`, `ed25519`, and `sha` jets) with our own Rust-native functions. Resolves #156.

TODO:
- [x] make sure none of the functions allocate (`alloc` features are disabled in all dependency crates)
- [x] `ed`
  - [x] `puck`
  - [x] `shar`
  - [x] `sign`
  - [x] `veri`
- [x] `aes`
  - [x] `siv{a,b,c}_en`
  - [x] `siv{a,b,c}_de`
- [x] replace `urcrypt` functions with `ares_crypto` in all jets
- [x] build and boot to `dojo` without `urcrypt`
- [x] ensure zero-length messages are handled correctly


rust/ares_crypto/src/ed25519.rs
```diff
@@ -1,38 +1,84 @@
...
+            ]
+        );
     }
-    Ok(())
 }
 
 #[cfg(test)]
 #[cfg(feature = "test_vs_urcrypt")]
-mod ucrypt_test {
-    // tests comparing urcrypt and ares_crypto output for ed25519 functions go here
```
eamsden: please leave this comment in and extend it with the cargo command to run these tests, since a regular `cargo test` will not run them

eamsden (changes requested):

rust/ares/src/noun.rs
```diff
@@ -249,6 +249,11 @@ impl DirectAtom {
         let bytes: &[u8; 8] = unsafe { std::mem::transmute(&self.0) };
         &bytes[..]
     }
+
+    pub fn as_mut_bytes(&mut self) -> &mut [u8] {
+        let bytes: &mut [u8; 8] = unsafe { std::mem::transmute(&mut self.0) };
+        &mut bytes[..]
+    }
```
eamsden, line 256: No.

- `transmute` is banned
- This will return a mutable reference to the existing direct atom, causing the atom to be modified by any writes through the mutable reference.

If this is being used to construct a direct atom by creating a local initialized to `D(0)` and mutating it, then there are a couple of alternatives:

1. Create a mutable local of type `[u8; 8]` and, when it is properly initialized, call `u64::from_le_bytes()` and initialize the atom from there
2. Create an indirect atom using `IndirectAtom::new_raw_mut_bytes()` and call `normalize_to_atom()` afterwards.

The second approach is always preferred because it ensures we do not accidentally construct a floating indirect atom or cell by overflowing the 63 bits of a direct atom.

eamsden (changes requested):
Actually looks pretty good, but I tagged a couple of issues.

rust/ares/src/noun.rs
```diff
@@ -806,6 +824,14 @@ impl Atom {
...
     }
 
+    pub fn as_mut_bytes(&mut self) -> &mut [u8] {
+        if self.is_direct() {
+            unsafe { self.direct.as_mut_bytes() }
+        } else {
+            unsafe { self.indirect.as_mut_bytes() }
+        }
+    }
+
```
eamsden, line 834: No. See above.

eamsden (changes requested):
Using `as_mut_bytes()` on atoms will cause these atoms to be mutated in place, breaking referential transparency. Please correct all instances of this.

matthew-levan:
> Using `as_mut_bytes()` on atoms will cause these atoms to be mutated in place, breaking referential transparency. Please correct all instances of this.

Done. Now working on ensuring zero-length messages are handled correctly.

matthew-levan:
@eamsden I ripped out the `as_mut_bytes` implementation from `DirectAtom` and replaced instances where it was used with new `IndirectAtom` allocations which I copy data into. I've also updated jets to handle zero-lengths correctly and added tests to demonstrate. This is ready for another review. The code isn't the cleanest, but I'm tagging you here for expediency nonetheless.

rust/ares/src/jets/lock/aes.rs
```diff
@@ -185,59 +159,42 @@ mod util {
...
+        key: &mut [u8; N],
         ads: Noun,
         txt: Atom,
-        fun: UrcryptSiv,
     ) -> Result {
         unsafe {
-            let siv_data = _allocate_atoms(stack, ads)?;
+            let ac_siv_data = _allocate_ads(stack, ads)?;
+            let siv_data: &mut [&mut [u8]] = std::slice::from_raw_parts_mut(
+                ac_siv_data.as_mut_ptr() as *mut &mut [u8],
```
eamsden, line 178: why is this doubly indirected?

rust/ares/src/noun.rs
```diff
@@ -474,6 +483,10 @@ impl IndirectAtom {
         unsafe { from_raw_parts(self.data_pointer() as *const u8, self.size() << 3) }
     }
 
+    pub fn as_mut_bytes(&mut self) -> &mut [u8] {
+        unsafe { from_raw_parts_mut(self.data_pointer_mut() as *mut u8, self.size() << 3) }
+    }
+
```
eamsden, line 489: this should not exist. If a noun has already been created then we should never mutate it. If we need to write into a mutable buffer to make a noun then allocate a new one with `IndirectAtom::new_raw_mut_bytes`

eamsden (changes requested):
No `as_mut_ptr()` for atoms!

rust/ares/src/jets/lock/aes.rs
matthew-levan: Good question, I'm not sure exactly what's going on here or why it works. How are the lengths of the byte buffers in `AcAesSivData` structs being translated to the lengths of the byte slices in `siv_data` here? By constructing the `siv_data` variable as we do here, are we getting a slice of byte slices whose elements are `[bytes, length]`? If that's the caes, I'm not sure that's desirable. The tests (taken from both IETF test cases and `dojo` examples) do pass, though.

matthew-levan:
`_allocate_ads` means "allocate associated datas", not "allocate atoms". I
clarify this to note that we aren't using `ac_siv_data.as_mut_ptr()` to get
a mutable pointer to an atom. Does that make sense?

eamsden:
> `_allocate_ads` means "allocate associated datas", not "allocate atoms". I clarify this to note that we aren't using `ac_siv_data.as_mut_ptr()` to get a mutable pointer to an atom. Does that make sense?


that makes sense, but the PR still adds an `as_mut_ptr()` method for indirect atoms which simply should not exist.

matthew-levan:
Hm maybe it did at one time but I've since removed it. The only
modification to `noun.rs` now is the addition of `IndirectAtom::
new_raw_mut_bytearray`.

rust/ares/src/jets/lock/sha.rs
```diff
@@ -119,6 +143,14 @@ mod tests {
...
         let c = &mut init_context();
 
+        let sam = T(&mut c.stack, &[D(1), D(0)]);
+        assert_jet_ubig(
+            c,
+            jet_shas,
+            sam,
+            ubig!(_0x4abac214e1e95fe0c60df79d09cbd05454a4cb958683e02318aa147f2a5e6d60),
+        );
+
```
ashelkovnykov, line 153: Do these tests have any significance? Or are they based on dojo calls from a ship running on Vere?

Theoretically, the NIST is supposed to have standardized test cases, but both of the links I've found are returning a 503...
[1](https://csrc.nist.gov/groups/ST/toolkit/examples.html)
[2](https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/SHA256.pdf)

rust/ares_crypto/src/ed25519.rs
```diff
@@ -0,0 +1,212 @@
...
+        // let seed_src = &ubig!(_0x9d61b19deffd5a60ba844af492ec2cc44449c5697b326919703bac031cae7f60)
+        //     .to_le_bytes();
+        // let mut seed: [u8; 32] = [0; 32];
+        // seed.copy_from_slice(seed_src);
+        // let mut signature: [u8; 64] = [0; 64];
+        // ac_ed_sign(&msg, &seed, &mut signature);
+        // let signature_src = &ubig!(_0xe5564300c360ac729086e2cc806e828a84877f1eb8e5d974d873e065224901555fb8821590a33bacc61e39701cf9b46bd25bf5f0595bbe24655141438e7a100b).to_le_bytes();
+        // let mut test_signature: [u8; 64] = [0; 64];
+        // test_signature.copy_from_slice(signature_src);
+        // assert_eq!(signature, test_signature);
```
ashelkovnykov, line 84: Cruft left over from getting `ares_crypto` to work?

rust/ares_crypto/src/ed25519.rs
```diff
@@ -0,0 +1,212 @@
...
+            [
+                59, 106, 39, 188, 206, 182, 164, 45, 98, 163, 168, 208, 42, 111, 13, 115, 101, 50,
+                21, 119, 29, 226, 67, 166, 58, 192, 72, 161, 139, 89, 218, 41
+            ]
+        );
+    }
+
+    // #[test]
+    // fn test_ed_shar() {
+    // }
```
ashelkovnykov, line 69: Cruft left over from getting `ares_crypto` to work?

rust/ares/src/jets/lock/ed.rs
```diff
@@ -142,6 +132,16 @@ mod tests {
...
+        let sam = A(
+            &mut c.stack,
+            &ubig!(_0x0),
+        );
+        let ret = A(
+            &mut c.stack,
+            &ubig!(_0x29da598ba148c03aa643e21d77153265730d6f2ad0a8a3622da4b6cebc276a3b),
+        );
+        assert_jet(c, jet_puck, sam, ret);
+
```
ashelkovnykov, line 144: Is there any significance to these tests, or are they just from the dojo of a ship running on Vere?

rust/ares/src/jets/lock/ed.rs
```diff
@@ -47,17 +44,17 @@ pub fn jet_shar(context: &mut Context, subject: Noun) -> Result {
...
-        let (_, secret) = IndirectAtom::new_raw_mut_bytes(stack, 32);
+        let public = &mut [0u8; 32];
+        let secret = &mut [0u8; 32];
 
         let pub_bytes = pub_key.as_bytes();
         let sec_bytes = sec_key.as_bytes();
 
         public[0..pub_bytes.len()].copy_from_slice(pub_bytes);
-        secret[0..sec_bytes.len()].copy_from_slice(sec_bytes);
+        secret[0..pub_bytes.len()].copy_from_slice(sec_bytes);
```
ashelkovnykov, line 54: copy-paste error

ashelkovnykov (changes requested):

rust/ares/src/jets/lock/aes.rs
```diff
@@ -377,6 +334,22 @@ mod tests {
...
+        }
+        fn context(s: &mut NockStack) -> Noun {
+            let sample = T(s, &[D(0), D(0)]);
+            T(s, &[D(0), sample, D(0)])
+        }
+
+        let siv = A(&mut c.stack, &ubig!(0x8fb4085a9b93662ab44f911e47e9ccd));
+        let res = T(&mut c.stack, &[siv, D(0), D(0x0)]);
+        assert_jet_in_door(c, jet_sivb_en, &[sample], &[context], res);
+
```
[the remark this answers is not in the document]
matthew-levan: It's just one I got from the `dojo`.

rust/ares/src/jets/lock/aes.rs
```diff
@@ -404,6 +377,22 @@ mod tests {
...
+        }
+        fn context(s: &mut NockStack) -> Noun {
+            let sample = T(s, &[D(0), D(0)]);
+            T(s, &[D(0), sample, D(0)])
+        }
+
+        let siv = A(&mut c.stack, &ubig!(0x2c6aabc5bb251140e221d70bfb31c519));
+        let res = T(&mut c.stack, &[siv, D(0), D(0x0)]);
+        assert_jet_in_door(c, jet_sivc_en, &[sample], &[context], res);
+
```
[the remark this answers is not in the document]
matthew-levan: Same answer as above.

rust/ares/src/jets/lock/ed.rs
matthew-levan: Not sure what you mean here. Will you clarify please?

rust/ares/src/jets/lock/ed.rs
matthew-levan: The significance is that they are tests from a real Vere `dojo`, yes. Future work to add test vectors from RFC 8032 is noted in the above comment.

rust/ares_crypto/src/sha.rs
```diff
@@ -0,0 +1,118 @@
+use sha1::{Digest, Sha1};
+use sha2::{Sha256, Sha512};
+
+/// Hashes a message using SHA-1.
+pub fn ac_sha1(message: &mut [u8], out: &mut [u8]) {
+    message.reverse();
+    let mut hasher = Sha1::new();
+    hasher.update(message);
+    let mut result = hasher.finalize();
```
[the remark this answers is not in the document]
matthew-levan: Nice, good catch.

rust/ares/src/noun.rs
```diff
@@ -436,6 +436,15 @@ impl IndirectAtom {
         (noun, from_raw_parts_mut(ptr as *mut u8, size))
     }
 
+    /// Create an indirect atom backed by a fixed-size array
+    pub unsafe fn new_raw_mut_bytearray<'a, const N: usize, A: NounAllocator>(
+        allocator: &mut A,
+    ) -> (Self, &'a mut [u8; N]) {
+        let word_size = (std::mem::size_of::<[u8; N]>() + 7) << 3;
```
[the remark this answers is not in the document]
matthew-levan: I think you're right, though-- take, for example, `N` is `8`. In the case of a left shift (as written), the result is `64`, not `1` as it should be.

rust/ares/src/jets/lock/ed.rs
matthew-levan: Ah, nevermind!

rust/ares_crypto/src/ed25519.rs
ashelkovnykov: Ah, nevermind - I see now. It's just a stub for a test that's yet unimplemented (unlike the other functions in this file, `shar` is currently only tested in relation to `urcrypt`).

ashelkovnykov:
This PR is almost ready to ship.

Minimum necessary changes:
- [byte array allocation size bug](https://github.com/urbit/ares/pull/189#discussion_r1472702826)
- [SHA message slicing bug](https://github.com/urbit/ares/pull/189#discussion_r1474046046)

Changes I'd prefer to have, but aren't necessary to merge:
- [singleton `hasher`s](https://github.com/urbit/ares/pull/189#discussion_r1472719317)
- Implement test cases from well known sources (e.g. RFCs, NIST, etc.). We can get away without these, methinks, since we're now relying on Rust crates which are very well tested - but better safe than sorry!

matthew-levan (commented):
Thanks for the review @ashelkovnykov. I pushed a couple commits to address the byte array allocation size bug and the SHA message slicing bug. With respect to the single `hasher` use in our SHA module, where would the singleton be initialized? A global? Passed in to the functions via the caller? I propose merging ASAP and circling back to use the singleton pattern afterwards.

rust/ares/src/jets/lock/aes.rs
matthew-levan: @ashelkovnykov would you mind taking a look at this as well?

rust/ares/src/jets/lock/sha.rs
```diff
@@ -17,19 +17,22 @@ pub fn jet_shas(context: &mut Context, subject: Noun) -> Result {
...
-            salt.len(),
-            message.as_ptr(),
-            message.len(),
-            out.as_mut_ptr(),
-        );
+
+        let msg_len = met(3, ruz);
+        if msg_len > 0 {
+            let (_msg_ida, message) = IndirectAtom::new_raw_mut_bytes(stack, msg_len);
+            message.copy_from_slice(&ruz.as_bytes()[0..msg_len]);
```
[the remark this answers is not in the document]
matthew-levan: We definitely can @ashelkovnykov, a pretty easy job too.

rust/ares/src/jets/lock/aes.rs
ashelkovnykov: This is correct:

The input is a list of atoms which are converted into a contiguous array of `AcAesSivData` objects, which is then used by the encoder. We pass it from the jets code as an array of byte arrays, because we need to reverse the byte ordering prior to encoding for each element. When used by the encoder, it implicitly converts our input to just a byte buffer. Vere does something similar.

We could make the byte ordering reversal explicit in the jet code and then pass just a byte buffer, if we think that's best. However, that's a different discussion than correctness.

ashelkovnykov (approved):

matthew-levan:
I removed `nix` from the CI in the latest commit, so it will need to be reverted and the CI error will need to be resolved. I wasn't able to fix the CI error quickly without removing `nix`.

ashelkovnykov:
@matthew-levan That's okay, @eamsden and I are going to look at fixing the Nix build using `musl`

PR #190, opened 2023-12-18
jets: jet +turn

eamsden:
Adds a jet for +turn, which caches the jet (or the non-existence of a jet) for the gate in the sample, and builds the list in destination-passing style.


frodwith:
Code that caches the call target for a gate should ensure that if the core is registered, +6 is irrelevant to matching. See https://github.com/urbit/vere/blob/develop/pkg/noun/jets.c#L1694 but note that this check is slightly pessimistic, and we should recheck after filling in each sample for maximum correctness.

frodwith:
one way to resturucture would be to select a function for slamming based on the jet match (always jet, never jet, check every time) and use that in the loop.

eamsden:
converting to draft until I can address @frodwith's comments

ashelkovnykov (commented):
Current code looks good, pending changes to implement @frodwith 's advice

philipquirk (approved):
lgtm

PR #193, opened 2023-12-19
jets: fix jet_rev

drbeefsupreme:
`jet_rev` had an issue with the indirect atom case. `IndirectAtom::new_raw` would normally be fine since we normalize, but since we're reversing endianness, we end up using the extra space. So the fix is to use `IndirectAtom::new_raw_bytes` instead.

I've also added a test case that catches this, which fails for the previous version.

### DIFF
--- a/rust/ares/src/jets/bits.rs
+++ b/rust/ares/src/jets/bits.rs
@@ -3,7 +3,7 @@
 use crate::interpreter::{Context, Error};
 use crate::jets::util::*;
 use crate::jets::{JetErr, Result};
-use crate::noun::{DirectAtom, IndirectAtom, Noun, D};
+use crate::noun::{IndirectAtom, Noun, D};
 use std::cmp;
 
 crate::gdb!();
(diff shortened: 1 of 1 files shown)


eamsden:
@drbeefsupreme this should actually call `IndirectAtom::new_raw_mut_bitslice()` and call `normalize_as_atom` on the indirect atom afterwards.

rust/ares/src/jets/bits.rs
```diff
@@ -226,7 +218,7 @@ pub fn jet_rev(context: &mut Context, subject: Noun) -> Result {
         dest[start..end].copy_from_bitslice(&src[(total_len - end)..(total_len - start)]);
```
eamsden, line 218: just checked and this isn't correct: +rev reverses *block* order, not bit order.

eamsden (changes requested):

rust/ares/src/jets/bits.rs
eamsden: ideally we have fastpaths for `boz == 0` (bitslice as above) and `boz >= 3` (byteslice) and correct-but-slow loop code for `boz == 1 || boz == 2`

drbeefsupreme: Isn't the reversal of block order being handled by the line above it? Otherwise I don't see how the test case I added could be passing (which I got from running `+rev` in dojo).

```
(rev 3 12 0xfaceb00c15deadbeef123456)
> 0x563412efbeadde150cb0cefa
```

I'm tired so maybe I'm misunderstanding. In any case, the fastpaths suggestion is great.

eamsden: you're right, diff-based review delenda est

eamsden (approved):

ashelkovnykov (approved):

PR #198, opened 2024-01-11
Implement fastpath for `jet_rev` where `boz = 0` and `boz = 3`

nisfeb:
```
Running benches/boz.rs (target/release/deps/boz-c09d868d611fc8e8)
boz 3                   time:   [20.110 ns 20.178 ns 20.282 ns]
                        change: [-3.3792% -3.1311% -2.8160%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/rust/ares/src/jets/bits.rs
+++ b/rust/ares/src/jets/bits.rs
@@ -6,6 +6,8 @@ use crate::jets::{JetErr, Result};
 use crate::noun::{IndirectAtom, Noun, D};
 use std::cmp;
 
+use self::util::lsh;
+
 crate::gdb!();
 
 /*
@@ -110,19 +112,8 @@ pub fn jet_end(context: &mut Context, subject: Noun) -> Result {
     let arg = slot(subject, 6)?;
     let (bloq, step) = bite(slot(arg, 2)?)?;
     let a = slot(arg, 3)?.as_atom()?;
-
-    if step == 0 {
-        Ok(D(0))
-    } else if step >= util::met(bloq, a) {
-        Ok(a.as_noun())
-    } else {
-        unsafe {
-            let (mut new_indirect, new_slice) =
-                IndirectAtom::new_raw_mut_bitslice(&mut context.stack, bite_to_word(bloq, step)?);
-            chop(bloq, 0, step, 0, new_slice, a.as_bitslice())?;
-            Ok(new_indirect.normalize_as_atom().as_noun())
-        }
-    }
+    let res = util::end(&mut context.stack, bloq, step, a)?;
+    Ok(res.as_noun())
 }
 
 pub fn jet_lsh(context: &mut Context, subject: Noun) -> Result {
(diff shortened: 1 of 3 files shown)


nisfeb:
Improved these cases with help from @eamsden. Based on our implementation, the tests may be incorrect. This also calls into question the correctness of the initial generic case. Either the tests or the cases need to be fixed. The original generic case has been left in place to help the investigation.

rust/ares/src/jets/bits.rs
```diff
@@ -200,25 +191,12 @@ pub fn jet_rev(context: &mut Context, subject: Noun) -> Result {
...
-
-    for (start, end) in (0..len).map(|b| (b << boz, (b + 1) << boz)) {
-        dest[start..end].copy_from_bitslice(&src[(total_len - end)..(total_len - start)]);
-    }
-
-    Ok(unsafe { output.normalize_as_atom() }.as_noun())
+    let mut dat = slot(arg, 7)?.as_atom()?;
+    dat = util::end(&mut context.stack, boz, len as usize, dat)?;
+    let dat_swp = util::swp(&mut context.stack, boz, dat);
+    let step = len as usize - util::met(boz, dat);
```
ashelkovnykov, line 198: This should be `met(boz, dat_swp)`, since a correct `swp` can return a smaller noun:
```
> `@ub`(swp 2 0b1111.1000.0000)
0b1000.1111
```

rust/ares/src/jets/bits.rs
```diff
@@ -304,6 +282,81 @@ pub mod util {
...
 
+    pub fn swp(stack: &mut NockStack, boz: usize, dat: Atom) -> Atom {
+        let len = met(boz, dat);
+        match boz {
+            0 => unsafe {
+                let (mut atom, dest) = IndirectAtom::new_raw_mut_bitslice(stack, (len + 63) >> 6);
+                dest.copy_from_bitslice(dat.as_bitslice());
+                dest.reverse();
+                atom.normalize_as_atom()
+            },
```
ashelkovnykov, line 293: This always returns at the granularity of one or more words of memory, which is not in accord with `++swp` in Arvo. For `0x87.6543`, this will return `0xc2a6e10000000000` (exactly 64-bits) and the code in `jet_rev` never right shifts.

rust/ares/src/jets/bits.rs
```diff
@@ -304,6 +282,81 @@ pub mod util {
...
+            3 => unsafe {
+                let (mut atom, dest) = IndirectAtom::new_raw_mut_bytes(stack, (len + 7) >> 3);
+                dest.copy_from_slice(dat.as_bytes());
```
ashelkovnykov, line 296: Slice bit length errors here, between `dest` and `dat`

ashelkovnykov (changes requested):
The tests are fine; the code has errors

eamsden:
I'm going to defer this until it comes up in benchmarking.

PR #201, opened 2024-01-16
PMA file extension

barter-simsum:
Makes the required changes to extend the PMA's backing file when allocation space is exceeded.

Includes a test in `btest.c` to validate this behavior. Other tests also exercise it (though don't perform file size validation because they were written to test different behavior)


eamsden (changes requested):
Lots of comments address the "frontier" which we decided we don't need. This is confusing, please clean those up.

barter-simsum:
> Lots of comments address the "frontier" which we decided we don't need. This is confusing, please clean those up.

done

eamsden (approved):

PR #204, opened 2024-01-23
pma: partition striping

barter-simsum:


eamsden:
confirmed with @barter-simsum that this is ready for review. @ashelkovnykov please review

rust/ares_pma/c-src/btree.c
```diff
@@ -1037,6 +1155,15 @@ _nlist_insert(BT_state *state, BT_nlistnode **dst, pgno_t nodepg)
...
+_nlist_insertn(BT_state *state, BT_nlistnode **dst, pgno_t lo, pgno_t hi)
+{
+  _nlist_insert2(state,
+                 dst,
+                 _node_get(state, lo),
+                 _node_get(state, hi));
+}
+
 static void
 _pending_nlist_merge(BT_state *state)
```
eamsden, line 1168: did we check that this doesn't have the same issue that was just uncovered in #216 for the flist?

eamsden (commented):
I need to think more about the comment at the end of the file. I think you're right and the btree node references can be offsets into the memory arena if we properly map the node page stripes, obviating the need to compute memory addresses by reference to the table (since the MMU is effectively doing that for us). But I need to mull it over to be sure.

rust/ares_pma/c-src/btree.c
barter-simsum: Indeed, it has the same issue. Will fix

barter-simsum:
> I need to think more about the comment at the end of the file. I think you're right and the btree node references can be offsets into the memory arena if we properly map the node page stripes, obviating the need to compute memory addresses by reference to the table (since the MMU is effectively doing that for us). But I need to mull it over to be sure.

I chose strategy (2) where a pgno for a node is a page offset into the memory arena, NOT a file offset into the persistent file. This works and as you pointed out, "obviating the need to compute memory addresses by reference to the table (since the MMU is effectively doing that for us)"

rust/ares_pma/c-src/btree.c
```diff
@@ -379,17 +393,56 @@ _node_get(BT_state *state, pgno_t pgno)
...
+ */
+#endif
   return FO2PA(state->map, pgno);
 }
 
 /* ;;: I don't think we should need this if _bt_nalloc also returns a disc offset */
 static pgno_t
 _fo_get(BT_state *state, BT_page *node)
 {
+  /* ;;: This may need to be fixed to accommodate partition striping */
```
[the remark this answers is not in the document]
barter-simsum: fixed

rust/ares_pma/c-src/btree.c
barter-simsum: fixed

rust/ares_pma/c-src/btree.c
```diff
@@ -379,70 +379,16 @@ struct BT_findpath {
...
-  - sector that contains node is i-1
-  - appropriately offset into i-1th fixed size partition: 2M, 8M, 16M, ...
-
-  */
-
-  /* for now, this works because the 2M sector is at the beginning of both the
-     memory arena and pma file
-  */
+  /* TODO: wrap in DEBUG macro a looping assert to confirm the given pgno is
+     contained within a node partition */
```
eamsden, line 383: Is there a reason not to turn this on all the time? It wouldn't be that slow, the node partition table is very small.

barter-simsum: oh duh, actually the pgno given is a page sized offset into the memory arena, so we can just check that it's less than the maximum extent of the last partition (since they're all mapped at the beginning).

barter-simsum: done

eamsden (commented):
These changes LGTM, please await @ashelkovnykov approval though.

rust/ares_pma/c-src/btree.c
```diff
@@ -1591,8 +1663,8 @@ _mlist_new(BT_state *state)
      page (alloced or not) in the persistent file. */
 static void
 _flist_grow(BT_state *state, size_t pages)
-/* grows the backing file by PMA_GROW_SIZE_p and appends this freespace to the
-   flist */
+/* grows the backing file by maximally `pages' or minimally PMA_GROW_SIZE_p and
```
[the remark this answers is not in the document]
barter-simsum: fixed

ashelkovnykov (commented):
I've been walking through the PR and PMA code trying to understand both at the same time. Since it looks like my review of this PR is the blocker, I just fast-forwarded through the PR.

It all *looks* fine, but that's all I'm confident in saying right now since I still don't fully understand the PMA (currently I'm going through the PMA files line by line and leaving comments for myself). One thing I'll note is that there's a bunch of "TODOs" still left, various redundancies, etc.

Overall, my feelings towards the PMA can be summed up as, "it's fine for alpha, assuming that we have tests that pass and that anytime we find a bug, we add it to the tests."

TL;DR - I don't see anything obviously wrong with the PR, but I've yet to convince myself that it's correct within the context of current PMA code. If that's enough for the team to merge to keep going, go ahead. I can add an explicit "approve" on GitHub as well, for posterity.

eamsden (approved):

PR #206, opened 2024-01-30
pma: fix file growth infinite loop

### LINKED ISSUE #205: PMA goes into infinite loop when flist has exactly one block and you allocate something of exactly that size
I'm getting an infinite loop in the PMA.  Here's how it happens:

- state->flist has exactly one block.
- It allocates something which is the exact size of this block (in my case the size is just 1 page).
- in _flist_record_alloc it hits the [last case](https://github.com/urbit/ares/blob/status/rust/ares_pma/c-src/btree.c#L520). head is state->flist since there's only node and head->next is NULL and it frees head and sets state->flist to NULL.
- Now the next time the PMA tries to find a free block in _bt_falloc it fails to find one and calls _flist_grow to add one.
- _flist_grow calls ftruncate to increase the file size and calls _flist_insert to add the new memory to the flist.
- _flist_insert does not check if *dst is NULL and so the while loop immediately exists and it doesn't add any block.
- _bt_falloc tries again to find a free block and fails, so it calls _flist_grow and _flist_insert ot add more space, which doesn't check if *dst is NULL, ...

I tried a couple fixes for this such as making _flist_insert check if *dst is NULL and if so just allocating a new block, or doing this inside _flist_grow. Both cause crashes so I don't know what the real solution is.

barter-simsum:
resolves #205

### DIFF
--- a/rust/ares_pma/c-src/btree.c
+++ b/rust/ares_pma/c-src/btree.c
@@ -1100,6 +1100,14 @@ _flist_insert(BT_flistnode **dst, pgno_t lo, pgno_t hi)
     *dst = new;
     return;
   }
+
+  /* otherwise, insert discontinuous node */
+  BT_flistnode *new = calloc(1, sizeof *new);
+  new->lo = lo;
+  new->hi = hi;
+  new->next = *dst;
+  *dst = new;
+  return;
 }
 
 static void
--- a/rust/ares_pma/c-src/btree.c
+++ b/rust/ares_pma/c-src/btree.c
@@ -733,6 +733,19 @@ _bt_dirtychild(BT_page *parent, size_t child_idx)
 }
 
 static int
+_bt_dirtydata(BT_page *leaf, size_t child_idx)
+/* effectively the same as _bt_dirtychild (setting the dirty bit at child_idx in
+   the given node), with the exception that we don't assert the dirty bit isn't
+   set. (Data may be written to the same fileoffset multiple times (a
+   malloc-free-malloc cycle) */
+{
+  assert(child_idx < 2048);
+  uint8_t *flag = &leaf->head.dirty[child_idx >> 3];
+  *flag |= 1 << (child_idx & 0x7);
+  return BT_SUCC;
+}
+
+static int
 _bt_cleanchild(BT_page *parent, size_t child_idx)
 {
   assert(_bt_ischilddirty(parent, child_idx));
@@ -1425,6 +1438,8 @@ _bt_insert2(BT_state *state, vaof_t lo, vaof_t hi, pgno_t fo,
 
   /* nullcond: node is a leaf */
   if (meta->depth == depth) {
+    /* dirty the data range */
+    _bt_dirtydata(node, childidx);
     /* guaranteed non-full and dirty by n-1 recursive call, so just insert */
     return _bt_insertdat(lo, hi, fo, node, childidx);
   }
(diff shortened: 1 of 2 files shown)


barter-simsum:
* not quite merge ready. this can potentially break `bt_free`. Freelist insertion needs to be revised

barter-simsum:
@eamsden @philipquirk I believe this is merge-ready now. The first test in btest.c has been disabled because it directly calls _bt_insert. When closing and syncing the btree, this causes errors because the ranges we've pretended to allocate (but haven't actually) aren't mmaped. Leaving disabled for now until I make test changes to override the sync procedure it calls (to avoid actually syncing).

rust/ares_pma/c-src/btree.c
```diff
@@ -2742,7 +2763,11 @@ bt_free(BT_state *state, void *lo, void *hi)
   BT_kv kv = leaf->datk[childidx];
   vaof_t offset = looff - kv.va;
   lopg = kv.fo + offset;
-  hipg = lopg + (looff - hioff);
+  hipg = lopg + (hioff - looff);
+
+  /* insert null into btree */
+  _bt_insert(state, looff, hioff, 0);
```
barter-simsum, line 2769: @eamsden Note, I've confirmed that bt_free is properly referencing the leaf's dirty bitmap (i.e. checking that the data ranges are actually dirty).

philipquirk:
This crashes for me with: 

ares(8146,0x1dce1dec0) malloc: *** error for object 0x600003198090: pointer being freed was not allocated

barter-simsum:
> This crashes for me with:
> 
> ares(8146,0x1dce1dec0) malloc: *** error for object 0x600003198090: pointer being freed was not allocated
> 
> It crashes the first time I call pma_sync()

@philipquirk  Can you give more context for the crash? When running it as a serf of vere, I get a crash, but not this one. For instance, can ycargo build --features "stop_for_debug"ou at least print the results of `info stack` after attaching to the ares process with gdb (build with the `"stop_for_debug"` feature).

philipquirk:
> > This crashes for me with:
> > ares(8146,0x1dce1dec0) malloc: *** error for object 0x600003198090: pointer being freed was not allocated
> > It crashes the first time I call pma_sync()
> 
> @philipquirk Can you give more context for the crash? When running it as a serf of vere, I get a crash, but not this one. For instance, can you at least print the results of `info stack` after attaching to the ares process with gdb (build with the `"stop_for_debug"` feature).

I just tried it again and it no longer crashes on launch for me. I'm not sure what's difference. I'm not seeing any crash at the moment.

philipquirk:
> > > This crashes for me with:
> > > ares(8146,0x1dce1dec0) malloc: *** error for object 0x600003198090: pointer being freed was not allocated
> > > It crashes the first time I call pma_sync()
> > 
> > 
> > @philipquirk Can you give more context for the crash? When running it as a serf of vere, I get a crash, but not this one. For instance, can you at least print the results of `info stack` after attaching to the ares process with gdb (build with the `"stop_for_debug"` feature).
> 
> I just tried it again and it no longer crashes on launch for me. I'm not sure what's difference. I'm not seeing any crash at the moment.

I just saw it crash when calling save in the serf (so probably in save_to_pma but i didn't attach a debugger)

barter-simsum:
@eamsden have confirmed this doesn't crash on latest commit from status branch. Ready for review

eamsden (approved):

PR #210, opened 2024-02-08
remove cargo-watch for darwin-x86 because it doesn't build

philipquirk:
The mac x86 build fails because the cargo-watch build fails with some weird build error inside the quartz framework headers. This fixes the build by excluding cargo-watch.

### DIFF
--- a/rust/flake.nix
+++ b/rust/flake.nix
@@ -29,14 +29,15 @@
               "rustfmt"
               "rust-src"
             ])
-            pkgs.cargo-watch
             pkgs.bacon
             pkgs.iconv
             pkgs.llvmPackages.clang
             pkgs.pkg-config
             pkgs.urcrypt
           ] ++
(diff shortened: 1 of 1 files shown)


rust/flake.nix
```diff
@@ -29,14 +29,15 @@
...
             pkgs.iconv
             pkgs.llvmPackages.clang
             pkgs.pkg-config
             pkgs.urcrypt
           ] ++
-            (nixpkgs.lib.lists.optional (parsedSystem.kernel.name != "darwin") pkgs.gdb); # nixpkgs won't build gdb for darwin
+            (nixpkgs.lib.lists.optional (parsedSystem.kernel.name != "darwin") pkgs.gdb) # nixpkgs won't build gdb for darwin
+            ++
+            (nixpkgs.lib.lists.optional (parsedSystem.kernel.name != "darwin") pkgs.cargo-watch); # nixpkgs won't build gdb for darwin
         };
```
philipquirk, line 41: This is wrong and should only exclude for darwin-x86 and not all of darwin but I don't know how to do that.

eamsden (approved):

PR #211, opened 2024-02-10
guard page / `bail:meme`

ashelkovnykov:
A simplified version of #202 that's ready for review, but still sprinkled with debugging console messages. Works with both `oom.pill` and `slim.pill`.

Inspired **very** heavily by @matthew-levan 's work on #202 and by our DMs regarding his troubleshooting efforts for it.


rust/ares/src/serf.rs
```diff
@@ -401,7 +402,28 @@ fn slam(context: &mut Context, axis: u64, ovo: Noun) -> Result<Noun, Error> {
     let sam = T(stack, &[D(6), D(0), D(7)]);
     let fol = T(stack, &[D(8), pul, D(9), D(2), D(10), sam, D(0), D(2)]);
     let sub = T(stack, &[arvo, ovo]);
-    interpret(&mut context.nock_context, sub, fol)
+
+    let frame_p = stack.get_frame_pointer();
+    let stack_pp = stack.get_stack_pointer_pointer();
+    let alloc_pp = stack.get_alloc_pointer_pointer();
```
joemfb, line 408: how does this mechanism of propagating nockstack bounds to the guard page signal handler account for stack frames pushed after the `call_with_guard()` invocation, but before the SIGSEGV?

ashelkovnykov: It takes as input the addresses of the stack and allocation pointers in the `NockStack` object, so as the program runs the values those pointers reference are automatically updated underneath it.

matthew-levan: The implementation as written here only provides the `guard` function visibility to the current frame's stack and heap boundary pointers, but not those of future frames that are pushed onto the `NockStack` afterwards by the `interpret` closure, if I understand correctly.

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+    return guard_oom;
+  }
+
+  uintptr_t old_guard_p = *guard_pp;
+  uintptr_t new_guard_p;
+  int32_t   err = 0;
+
+  fprintf(stderr, "guard: focus: old guard = %p\r\n", (void *)old_guard_p);
+
+  // Unmark the old guard page (if there is one)
```
matthew-levan, line 82: We should enforce the invariant that there will always be a guard page installed. This code uninstalls the existing (old) guard page before installing a new one, which violates that invariant.

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+      fprintf(stderr, "guard: focus: unmark error\r\n");
+      return err;
+    }
+  }
+
+  // Compute new guard page
+  // XX:  Should we also check for new_guard_p < min(stack_p, alloc_p)?
+  new_guard_p = GD_PAGE_ROUND_DOWN((stack_p + alloc_p) / 2);
+  fprintf(stderr, "guard: focus: new guard = %p\r\n", (void *)new_guard_p);
+  if (new_guard_p == old_guard_p) {
```
matthew-levan, line 94: Checking if `new == old` for detecting memory exhaustion will trigger a false positive when a nested `call_with_guard` call happens to have stack and allocation pointers which are equidistant from the old guard page.

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+    fprintf(stderr, "guard: prot: mprotect error %d\r\n", errno);
+    fprintf(stderr, "%s\r\n", strerror(errno));
+    return guard_mprotect | errno;
+  }
+
+  return 0;
+}
+
+static int32_t
+_mark_page(void *address)
```
matthew-levan, line 45: Respectfully, this is cruft. There's no need for `_prot_page` or `_mark_page`-- adds unnecessary complexity/lines without much gain.

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+/**
+ *  XX: documentation
+ */
+typedef struct _gs {
+  uintptr_t         guard_p;
+  const uintptr_t  *stack_pp;
+  const uintptr_t  *alloc_pp;
+  jmp_buf           env_buffer;
+  struct sigaction  prev_sa;
+} GuardState;
```
matthew-levan, line 27: Why not just use top-level `static` variables instead of having to thread a `malloc`'d needle through the various functions?

rust/ares/src/serf.rs
```diff
@@ -401,7 +402,28 @@ fn slam(context: &mut Context, axis: u64, ovo: Noun) -> Result<Noun, Error> {
...
+
+    let res = call_with_guard(
+        stack_pp as *const *const u64,
+        alloc_pp as *const *const u64,
+        &mut || interpret(&mut context.nock_context, sub, fol),
+    );
+
+    if let Err(Error::NonDeterministic(Mote::Meme, _)) = res {
```
matthew-levan, line 416: I think this frame unwinding procedure is one of the missing links in #202.

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+_signal_handler(int sig, siginfo_t *si, void *unused)
+{
+  uintptr_t sig_addr;
+  int32_t   err = 0;
+
+  fprintf(stderr, "guard: sig_handle: %d received\r\n", sig);
+
+  if (sig != SIGSEGV) {
+    fprintf(stderr, "guard: sig_handle: invalid signal\r\n");
+    siglongjmp(_guard_state->env_buffer, guard_signal);
```
matthew-levan, line 122: Should we long jump for *any* signal other than `SIGSEGV` here, or just `SIGINT`?

rust/ares_guard/c-src/guard.c
```diff
@@ -0,0 +1,287 @@
...
+    struct sigaction prev_sa = _guard_state->prev_sa;
+
+    fprintf(stderr, "guard: page at %p miss\r\n", (void *)_guard_state->guard_p);
+
+    if (prev_sa.sa_sigaction != NULL) {
+      prev_sa.sa_sigaction(sig, si, unused);
+    } else if (prev_sa.sa_handler != NULL) {
+      prev_sa.sa_handler(sig);
+    } else {
+      // There should always be a default SIGSEGV handler
```
matthew-levan, line 152: Who installs the default `SIGSEGV` handler? The OS?

rust/ares/src/serf.rs
ashelkovnykov: That's incorrect - there's only one `NockStack` object that exists on the heap for the entire lifetime of Ares as Serf. If we take the address of the `stack_pointer` and `alloc_pointer` members, they will always be up-to-date when we de-reference the pointers.

rust/ares_guard/c-src/guard.c
ashelkovnykov: I don't think this actually matters, because only one thread is ever allowed to write to the `NockStack` and `SIGSEGV` is added to the signal mask, meaning that our `SIGSEGV` signal handler cannot be interrupted by another `SIGSEGV`. No read-only thread should hit the guard page, since there's no data there yet (by design). Even if it does (somehow), it'll hang in place until our signal handler finishes resetting the guard page, and then attempt to repeat the instruction, which will now work.

rust/ares_guard/c-src/guard.c
ashelkovnykov: That's true, though we're protected by two invariants:
1) In this version of the code, the invariant is that we can only have one call at a time
2) In the future version of the code, the invariant should be that the re-centering code is only ever called from the signal handler, unless there is no guard page currently configured

Both of these invariants should be documented.

rust/ares_guard/c-src/guard.c
ashelkovnykov: I'm trying to avoid having to c/p the `return guard_mprotect | errno;` at the site of every `mprotect()` call, though I'm open to getting rid of these 3 helpers.

rust/ares_guard/c-src/guard.c
ashelkovnykov: It's idiomatic in C that if you absolutely MUST have global state, all of the state should be wrapped in a struct that is:
- passed between functions as input
- `malloc`ed on first usage
- explicitly `free`ed on completion

rust/ares_guard/c-src/guard.c
ashelkovnykov: Any signal - something has gone fatally wrong if our signal handler was registered to handle **only** `SIGSEGV` signals, but has received a different signal.

I think the more appropriate question here is: should we even bother jumping back to the Rust code and `panic!`ing there, or should we just `assert(0)` immediately. @eamsden @joemfb @barter-simsum @frodwith ?

rust/ares_guard/c-src/guard.c
ashelkovnykov: That's correct - the OS installs a default `SIGSEGV` handler that logs the error and `abort()`s.

rust/ares_guard/c-src/guard.c
eamsden: assert(0) immediately if assumptions are broken in C code.

eamsden (approved):

PR #213, opened 2024-02-27
add format-to-cord implementation and flog! macro

eamsden:
This adds the ability to do debugging printfs, formatted using Rust format strings, by allocating outputs as cords on the Nockstack and sending them as %flog messages via Newt


eamsden:
(cherry-picked from codegen branch)

matthew-levan (approved):
Tested this with a fresh fakezod boot and a `flog!` call in `interpret`, works fine.

PR #216, opened 2024-03-12
pma: clear pending_flist after completion of _pending_flist_merge

barter-simsum:
use-after-free was responsible for crash


eamsden (approved):
Good find

PR #217, opened 2024-03-12
interpreter: check for copying flag in exit

eamsden:
An OOM nondeterministic error can cause exit() to begin with a frame in copying mode. This causes an assertion failure when we allocate a cell for the stack trace. 

This PR does not yet handle recursive OOMing (i.e. OOMing when we are making a stack trace) which should be handled, but it does handle the case of OOMing during a `preserve()`.

### DIFF
--- a/rust/ares/src/interpreter.rs
+++ b/rust/ares/src/interpreter.rs
@@ -1185,6 +1185,11 @@ fn exit(
     unsafe {
         context.restore(snapshot);
 
+        if context.stack.copying() {
+            assert!(context.stack.get_frame_pointer() != virtual_frame);
+            context.stack.frame_pop();
+        }
+
         let stack = &mut context.stack;
         let mut preserve = match error {
             Error::ScryBlocked(path) => path,
@@ -1192,6 +1197,7 @@ fn exit(
                 // Return $tang of traces
                 let h = *(stack.local_noun_pointer(0));
                 // XX: Small chance of clobbering something important after OOM?
+                // XX: what if we OOM while making a stack trace
                 T(stack, &[h, t])
             }
         };
--- a/rust/ares/src/mem.rs
+++ b/rust/ares/src/mem.rs
@@ -137,6 +137,10 @@ impl NockStack {
         };
     }
 
+    pub fn copying(&self) -> bool {
+        self.pc
+    }
+
     /** Current frame pointer of this NockStack */
     pub fn get_frame_pointer(&self) -> *const u64 {
         self.frame_pointer


philipquirk:
This fixed the memory crash I was seeing.

rust/ares/src/interpreter.rs
```diff
@@ -1185,13 +1185,19 @@ fn exit(
     unsafe {
         context.restore(snapshot);
 
+        if context.stack.copying() {
+            assert!(context.stack.get_frame_pointer() != virtual_frame);
+            context.stack.frame_pop();
+        }
```
ashelkovnykov, line 1191: Hmmm... This is an interesting case. In copying mode, we're not actually making any new data. Theoretically, there could be an ordering in which we can copy the data that is strictly not an OOM, since we will only overwrite already-copied data, and should therefore be allowed to ignore the guard page. However, there's no guarantee of this (picture a single `N`-page atom allocated into open `NockStack` space of size less than `2*N`). Furthermore, we have no control over the order in which atoms are copied during `preserve`.

It's possible that we could avoid more of these cases if we changed the guard page to `PROT_READ`, but not guaranteed. I imagine most OOM cases are such that we're going to hit them no matter what - it's just bad luck of where exactly in the loop cycle we hit them.

rust/ares/src/interpreter.rs
```diff
@@ -1185,13 +1185,19 @@ fn exit(
...
+
         let stack = &mut context.stack;
         let mut preserve = match error {
             Error::ScryBlocked(path) => path,
             Error::Deterministic(_, t) | Error::NonDeterministic(_, t) | Error::ScryCrashed(t) => {
                 // Return $tang of traces
                 let h = *(stack.local_noun_pointer(0));
                 // XX: Small chance of clobbering something important after OOM?
+                // XX: what if we OOM while making a stack trace
```
ashelkovnykov, line 1200: It's unconfirmed whether the true `mook` code still has bugs in it. If it doesn't, then we should uncomment the true code, which calls `mink`, and therefore also has OOM detection.

rust/ares/src/interpreter.rs
```diff
@@ -1185,13 +1185,19 @@ fn exit(
...
                 T(stack, &[h, t])
```
ashelkovnykov, line 1201: I think the real issue here is that we always move the `alloc` / `stack` pointers, and then rely on the guard page. If the pointers didn't move, or were moved back, then we could safely ignore the guard page because we can trust that since the guard page is an entire page, we certainly have room for one more `Cell` that references two existing `Noun`s.

Right now, the bug here is that the `alloc` and `stack` pointers might be way past each other (i.e. on the wrong side for their polarity) and writing this `Cell` clobbers something important. It also goes undetected, because we potentially don't touch the guard page, so we don't trigger another OOM in the parent `interpret` block (even if we did, there is no parent `interpret` for the root `interpret`).

rust/ares/src/interpreter.rs
eamsden: This is the core of the PR. The other stuff is just notation for future reference.

This fixes the bug (OOM during copying causes an assertion failure) in a minimally invasive way. It throws away stack traces but arguably we are done with that trace anyway since we were just doing preservation prior to a pop.

ashelkovnykov (approved):

PR #224, opened 2024-03-28
pma: fix regression on restore

barter-simsum:
There was a regression introduced with partition striping that caused problems on restore due to node partitions incidentally getting mapped anonymous rather than to the backing file


barter-simsum:
* Will fix cargo build warnings and clean up the code a bit

PR #226, opened 2024-04-12
pma: fix memory corruption restoration bug

barter-simsum:
Previously, we failed to record node partitions in the flist on restore - leading to memory corruption. No we do so.

btest.c for some reason ran without fault despite this bug. But when running ares as a serf of vere, this would routinely cause crashes in restoration. I've confirmed this PR fixes that and restoration seems to be behaving normally.


barter-simsum:
* I may refactor this a bit as the dependency graph on restore is pretty complicated and could probably be simplified. Perhaps in a different PR

barter-simsum:
PS. Please don't merge yet. Confirming this fix doesn't introduce any memory leaks in the backing file.

ashelkovnykov (approved):
LGTM - reviewed with @barter-simsum in pair session

PR #227, opened 2024-04-23
bump hoon root to %k.138

drbeefsupreme:
Not sure if there was a particular reason this hadn't been updated, or if nobody had gotten around to doing it yet.

I've confirmed that Nockchain runs fine with this. However, `test_sivb_en` fails. It currently passes on the `status` branch.

``` 
test jets::lock::aes::tests::test_siva_de ... ok
test jets::lock::aes::tests::test_siva_en ... ok
test jets::lock::aes::tests::test_sivb_de ... ok
memory allocation of test jets::lock::aes::tests::test_sivb_en ... 24ok bytes failed
```

I spent a few minutes looking into this but nothing popped out at me.

### DIFF
--- a/rust/ares/src/jets/hot.rs
+++ b/rust/ares/src/jets/hot.rs
@@ -4,9 +4,9 @@ use ares_macros::tas;
 use either::Either::{self, Left, Right};
 use std::ptr::{copy_nonoverlapping, null_mut};
 
-/** Root for Hoon %k.139
+/** Root for Hoon %k.138
  */
-pub const K_139: Either<&[u8], (u64, u64)> = Right((tas!(b"k"), 139));
+pub const K_138: Either<&[u8], (u64, u64)> = Right((tas!(b"k"), 138));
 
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
Hmmm, the test passed on CI. I wonder what's causing it to fail on my machine?

eamsden:
Rather than mutating, this should be copied, or perhaps we should write a const function that takes the kelvin root.

rust/ares/src/jets/hot.rs
```diff
@@ -4,9 +4,9 @@ use ares_macros::tas;
 use either::Either::{self, Left, Right};
 use std::ptr::{copy_nonoverlapping, null_mut};
 
-/** Root for Hoon %k.139
+/** Root for Hoon %k.138
```
eamsden, line 7: We need both roots in place, which means the hot state needs to be duplicated. A const function is probably the way here.

eamsden (changes requested):

PR #236, opened 2024-05-30
fix two weird bugs

philipquirk:
This fixes two weird bugs on mac:

1) In a debug build, copy_nonoverlapping with a NULL source and a 0 len will crash with an assertion. This possibly only happens on mac. Fix is to put an if around it to only call if len > 0.
2) Bitshifts that shift all the way to 0 are failing with an overflow assert. The assert is "attempt to shift right with overflow". Breaking it up into the shift - 1 and then a shift of 1 avoids this. This makes zero sense. I originally thought it was a bug introduced in the nightly toolchain but I can reproduce it in stable. Now i think its always been there but its very hard to activate- you have to have the bitmap in the PMA be exactly 0x800.0000. This causes next_chunk=trailing zeros to be 31 and so it shifts it by next_chunk+1=32, so all the way to 0.

### DIFF
--- a/rust/ares/src/hamt.rs
+++ b/rust/ares/src/hamt.rs
@@ -345,18 +345,22 @@ impl<T: Copy + Preserve> Hamt<T> {
                         *new_leaf_buffer = (*n, t);
                         let split = stem.hypothetical_index(chunk);
                         let new_buffer = stack.struct_alloc(stem.size() + 1);
-                        copy_nonoverlapping(stem.buffer, new_buffer, split);
+                        if split > 0 {
+                            copy_nonoverlapping(stem.buffer, new_buffer, split);
+                        }
                         *new_buffer.add(split) = Entry {
                             leaf: Leaf {
                                 len: 1,
                                 buffer: new_leaf_buffer,
                             },
                         };
-                        copy_nonoverlapping(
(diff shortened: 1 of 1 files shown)


rust/ares/src/hamt.rs
```diff
@@ -707,8 +713,10 @@ impl<T: Copy + Persist> Persist for Hamt<T> {
             let next_type = traversal[depth].typemap & (1 << next_chunk) != 0;
             let next_entry_ptr = traversal[depth].buffer;
 
-            traversal[depth].bitmap >>= next_chunk + 1;
-            traversal[depth].typemap >>= next_chunk + 1;
+            traversal[depth].bitmap >>= next_chunk;
+            traversal[depth].bitmap >>= 1;
+            traversal[depth].typemap >>= next_chunk;
+            traversal[depth].typemap >>= 1;
```
eamsden, line 719: Please turn this into an `if` that 0s bitmap and typemap unless `next_chunk > 31` and add a generous comment explaining the edge case

eamsden (changes requested):

rust/ares/src/hamt.rs
philipquirk: done

eamsden (approved):

PR #242, opened 2024-06-14
site: call site caching

matthew-levan:
This PR adds a very barebones call site caching module, and uses it for these jets:
- [x] `+turn` (updated implementation)
- [x] `+roll` (new jet)
- [x] `+rep:by` (new jet)
- [x] `+rep:in` (new jet; same as `+rep:by`)

Checkmarks denote successful testing on `macos-aarch64`.

Note that one of the big open PRs right now (I believe a codegen one) includes map/set jets in the `seam` module, but I called my map jets `maps`. Happy to reconcile later and add mine to `seam`, but not worth the git cherrypicking right now.


rust/ares/src/jets/hot.rs
```diff
@@ -219,6 +224,11 @@ pub const URBIT_HOT_STATE: &[HotEntry] = &[
         1,
         jet_jam,
     ),
+    (
+        &[K_139, Left(b"one"), Left(b"two"), Left(b"by"), Left(b"rep")],
+        1,
+        jet_by_rep,
+    ),
```
joemfb, line 231: Note, you can declare this same jet as `+rep:in`, the signatures are completely compatible.

matthew-levan:
Indeed, I will once I ensure it works correctly.

joemfb:
#243 should be investigated along with these changes.

Issue #246, opened 2024-07-03
jets::bits::util::bex() overallocates by 8x


eamsden:
new_raw_mut_bitslice expects size in words, bex computes size in bytes

PR #248, opened 2024-07-06
jets: fix bex overallocation

barter-simsum:
fixes: #246

### DIFF
--- a/rust/ares/src/jets/bits.rs
+++ b/rust/ares/src/jets/bits.rs
@@ -310,7 +310,7 @@ pub mod util {
             if arg < 63 {
                 DirectAtom::new_unchecked(1 << arg).as_atom()
             } else {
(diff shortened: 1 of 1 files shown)


rust/ares/src/jets/bits.rs
```diff
@@ -310,7 +310,7 @@ pub mod util {
             if arg < 63 {
                 DirectAtom::new_unchecked(1 << arg).as_atom()
             } else {
-                let (mut atom, dest) = IndirectAtom::new_raw_mut_bitslice(stack, (arg + 7) >> 3);
+                let (mut atom, dest) = IndirectAtom::new_raw_mut_bitslice(stack, (arg + 7) >> 5);
```
joemfb, line 313: The old code was correct for bytes. This new code wouldn't be quite right for 32 bit words, but Ares is using 64 bit words.

32-bit: `(arg + 31) >> 5`
64-bit: `(arg + 63) >> 6`

PR #267, opened 2024-09-14
snip and sew jets

litlep-nibbyt:


rust/sword/Cargo.toml
```diff
@@ -12,7 +12,7 @@ sword_guard = { path = "../sword_guard" }
 sword_crypto = { path = "../sword_crypto" }
 sword_macros = { path = "../sword_macros" }
 sword_pma = { path = "../sword_pma" }
-assert_no_alloc = { path = "../assert_no_alloc" }
+assert_no_alloc = { path = "../assert_no_alloc", features=["warn_debug"]}
```
philipquirk, line 15: i'm not sure exactly what setting for this should be merged

rust/sword/src/interpreter.rs
```diff
@@ -1657,9 +1657,10 @@ mod hint {
                             Err(cold::Error::NoParent) => {
                                 //  XX: Need better message in slog; need better slogging tools
                                 //      format!("could not find parent battery at given axis: {} {}", chum, parent_formula_ax)
+                                let chum = clue.slot(2).ok()?;
                                 let tape = tape(
                                     stack,
-                                    "serf: cold: register: could not find parent battery at given axis",
+                                    &format!("serf: cold: register: could not find parent battery at given axis: {}", chum),
```
philipquirk, line 1663: i really wish we could do this but format! allocates so it is forbidden. i constantly put this code in locally though.

philipquirk (changes requested):
Looks good except for calling format!

rust/sword/src/interpreter.rs
```diff
@@ -1659,7 +1659,7 @@ mod hint {
                                 //      format!("could not find parent battery at given axis: {} {}", chum, parent_formula_ax)
                                 let tape = tape(
                                     stack,
-                                    "serf: cold: register: could not find parent battery at given axis",
+                                    &format!("serf: cold: register: could not find parent battery at given axis"),
```
philipquirk, line 1662: you can just pass a string in here

philipquirk: I don't know if format allocates if it only has a string but why not just pass a string

philipquirk (approved):
I'm approving but i think you need to remove format

PR #269, opened 2024-09-18
snag and roll jets

litlep-nibbyt:
use flog for string interpolation to print out missing jet information.

### DIFF
--- a/rust/sword/src/site.rs
+++ b/rust/sword/src/site.rs
@@ -4,8 +4,11 @@ use bitvec::slice::BitSlice;
 
 use crate::interpreter::{interpret, Context};
 use crate::jets::util::slot;
-use crate::jets::Jet;
+use crate::jets::{Jet, JetErr};
 use crate::noun::{Noun, D, T};
+
+/// Return Err if the computation crashed or should punt to Nock
+pub type Result = std::result::Result<Noun, JetErr>;
 
 pub struct Site {
     pub battery: Noun,    // battery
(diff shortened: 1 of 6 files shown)


rust/sword/src/site.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct Site {
+    pub battery: Noun,    // battery
+    pub context: Noun,    // context
+    pub jet: Option<Jet>, // jet driver
+    pub path: Noun,       // label
+}
+
+impl Site {
+    /// Prepare a locally cached gate to call repeatedly.
+    pub fn new(ctx: &mut Context, core: &mut Noun) -> Site {
```
philipquirk, line 22: why is core mut? is this necessary?

philipquirk (approved):
This is great. I have one question which is why is core mut inside a Site? I assume its necessary but I don't see why it would ever be changed.

rust/sword/src/site.rs
eamsden: Yes unifying equality pollutes a lot of stuff with `&mut`

rust/sword/src/site.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Slam a cached call site.
+pub fn site_slam(ctx: &mut Context, site: &Site, sample: Noun) -> Result {
+    let subject = T(&mut ctx.stack, &[site.battery, sample, site.context]);
+    if site.jet.is_some() {
+        let jet = site.jet.unwrap();
+        jet(ctx, subject)
+    } else {
+        Ok(interpret(ctx, subject, site.battery)?)
+    }
+}
```
eamsden, line 73: This would be better as an impl method on the `Site` struct. But that need not block merge for now.

eamsden (approved):

PR #271, opened 2024-10-04
Hamt Iterator

bitemyapp:
image

First pass at the Hamt iterator. My understanding was we didn't need this for `MutHamt` so I didn't bother there. Max depth is 6 AIUI so the traversal stack is a fixed array.

### DIFF
--- a/rust/sword/src/hamt.rs
+++ b/rust/sword/src/hamt.rs
@@ -853,7 +853,8 @@ impl <'a, T: Copy>Iterator for Hamsterator<'a, T> {
                 Some((Right(leaf), _)) => {
                     // Found a leaf, return its value and prepare for next
                     self.traversal_stack[self.depth].1 += 1;
-                    if let Some(pair) = unsafe { leaf.to_mut_slice().get(0) } {
+                    let slice = unsafe { leaf.to_mut_slice() };
+                    if let Some(pair) = slice.get(0) {
                         return Some(*pair);
                     }
                 }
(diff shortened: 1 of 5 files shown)


rust/sword/src/hamt.rs
```diff
@@ -795,3 +800,146 @@ impl<T: Copy + Persist> Persist for Hamt<T> {
...
+    }
+}
+
+impl <'a, T: Copy>Iterator for Hamsterator<'a, T> {
+    type Item = (Noun, T);
+
+    // Iterate over the values in the HAMT
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if self.depth == usize::MAX {
```
drbeefsupreme, line 834: This seems like an unusual condition to exit a loop. It looks like it works fine, but is there a reason to use this over checking if `depth == 0` and returning `None` on line 840 instead?

rust/sword/src/hamt.rs
```diff
@@ -795,3 +800,146 @@ impl<T: Copy + Persist> Persist for Hamt<T> {
...
+                    self.traversal_stack[self.depth].1 += 1;
+                    self.depth += 1;
+                    self.traversal_stack[self.depth] = (next_stem, 0);
+                }
+                Some((Right(leaf), _)) => {
+                    // Found a leaf, return its value and prepare for next
+                    self.traversal_stack[self.depth].1 += 1;
+                    if let Some(pair) = unsafe { leaf.to_mut_slice().get(0) } {
+                        return Some(*pair);
+                    }
```
drbeefsupreme, line 859: A `Leaf` can have more than one key-value pair, see L414-433. Does this handle that case? It looks like it only handles the single pair case.

drbeefsupreme (changes requested):
One possible bug, one style issue

rust/sword/src/hamt.rs
bitemyapp: It's using `.wrapping_sub(1)` to detect when it's done. I realize this looks odd but rejiggering it to use a zero-based "done" is more annoying than you'd guess. Depth starts at zero, it never iterates if you change it to `== 0` with no other changes.

rust/sword/src/hamt.rs
bitemyapp: Let me construct a regression test for this so that I know when I've actually fixed it. Thank you!

drbeefsupreme (approved):

drbeefsupreme:
LGTM

rust/sword/src/hamt.rs
```diff
@@ -944,51 +952,89 @@ mod test {
...
+                    println!("FOUND HASH COLLISION!!!!! {} {} {} {}", x, y, n_hash, t_hash);
+                    let end = std::time::Instant::now();
+                    println!("Time: {:?}", end - start);
+                }
+                if y % 100000000 == 0 {
+                    println!("{} {}", x, y);
+                }
+            }
+        }
+    }
```
drbeefsupreme, line 1039: Comment this out again?

rust/sword/src/noun.rs
```diff
@@ -1042,6 +1042,12 @@ pub union Noun {
     allocated: Allocated,
 }
 
+impl Default for Noun {
+    fn default() -> Self {
+        Noun { raw: u64::MAX }
```
drbeefsupreme, line 1047: I do not want to merge this without @eamsden taking a look at this.

drbeefsupreme (commented):
Latest commit looks good except for `default()` which I defer to @eamsden

rust/sword/src/hamt.rs
```diff
@@ -421,7 +421,7 @@ impl<T: Copy + Preserve> Hamt<T> {
                             copy_nonoverlapping(stem.buffer, new_buffer, stem.size());
                             *new_buffer.add(idx) = Entry {
                                 leaf: Leaf {
-                                    len: leaf.len,
+                                    len: leaf.len + 1,
```
eamsden, line 426: OOF

Good catch boys @bitemyapp @drbeefsupreme

eamsden (approved):

PR #280, opened 2024-10-30
Draft: dropping sword guard

bitemyapp:


eamsden:
Work incorporated into #283

